PR #6, opened 2015-03-23
author update

vvo:
Of course you are the author, but WE @algolia are the authors :)

Same author as for the JS client, what do you think?


bobylito:
Resistance is futile ;)

Issue #25, opened 2015-03-27
Clear refinements matching a specific criteria


vvo:
Something like:

```
helper.clearRefinements(function(facet, value) {
  return facet === "foo" && value.indexOf("bar") === 0;
});
```

imported from https://github.com/algolia/algoliasearch-client-js/issues/33

bobylito:
It's definitely powerful to provide a function for filtering which parameters needs to be cleared. I'm wondering if we could something more to the point like : 

``` javascript
helper.clearRefinements( myFacetName );
```

Also we should think about the facet not all refinements are the same, and we may want to also insights on the type of refinement we want to clear. 

ping @redox

redox:
Using a single `facetName` parameter is definitely a feature we want. Being able to pass a function as well is to handle the hierarchical refinements. For instance, to remove all refinements of a specific level we could do something like:

```
helper.clearRefinements(function(facet, value) {
  return facet === "categories" && value.indexOf("Level 1/") === 0;
});
```

bobylito:
Facet refinement should be totally managed and shouldn't rely on people implementing this kind of function imho.
Ok for both the simple (facet name) and advanced (function) way of clearing though.

Jerskouille:
Adding `function (facet, value)` won't be that trivial, because of the `numericRefinements` which are composed of `name`, `operator` and `value`.
We might imagine merging `operator` and `value` ?

chrisyeung1121:
Just wondering if the problem I am facing is the one described in this issue?

I have a webapp that allows user to filter by `>=` and `<=` of the `daily_rate`.

`helper.addNumericRefinement('daily_rate','>=', $(this).val());`
`helper.addNumericRefinement('daily_rate','<=', $(this).val());`

And I want to be able to do `helper.clearRefinement('daily_rate','<=')`

Can I do that?

PR #50, opened 2015-04-14
Clear refinements by name

Jerskouille:
Designed to provide a first answer to #25.
Added 4 internal functions :
- clearFacetRefinements
- clearDisjunctiveFacetRefinements
- clearExcludeRefinements
- clearNumericRefinements

Each of them can clear everything but also be limitted by `name`.
It doesn't change how `helper.clearRefinements()` works, but
adds `helper.clearRefinements( name )` will search and remove
the refinements of a given facet / attribute.

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -3,6 +3,8 @@ var keys = require( "lodash/object/keys" );
 var forEach = require( "lodash/collection/forEach" );
 var reduce = require( "lodash/collection/reduce" );
 var isEmpty = require( "lodash/lang/isEmpty" );
+var isUndefined = require( "lodash/lang/isUndefined" );
+var isString = require( "lodash/lang/isString" );
 
 /**
  * @typedef FacetList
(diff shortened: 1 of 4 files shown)


src/SearchParameters/index.js
```diff
@@ -103,13 +105,19 @@ var SearchParameters = function( newParameters ) {
...
+   * @method
+   * @param {string} [name] - If given, name of the facet / attribute on which  we want to remove all refinements
+   * @return {AlgoliaSearchHelper}
+   */
+  clearRefinements : function clearRefinements( name ) {
+    return this
+      .clearNumericRefinements( name )
+      .clearFacetRefinements( name )
+      .clearExcludeRefinements( name )
+      .clearDisjunctiveFacetRefinements( name );
```
bobylito: Even though it's a good idea to reuse code, I wouldn't do that. mutateMe has a limitation, it always create a new instance, and because it is used in all those 4 methods, it will create sequentially 4 instances. It would be nice to revert that imho.

test/spec/helper.clears.js
```diff
@@ -0,0 +1,104 @@
...
+
+  t.deepEqual( helper.state.numericRefinements.numeric1, { ">=" : "0", "<" : "10" } );
+  helper.clearRefinements( "numeric1" );
+  t.assert( isUndefined( helper.state.numericRefinements.numeric1 ) );
+  t.doesNotThrow( function() {
+    helper.clearRefinements( "numeric1" );
+  } );
+
+  t.end();
+} );
```
bobylito: Good tests!!! :)

src/SearchParameters/index.js
Jerskouille: Done in https://github.com/algolia/algoliasearch-helper-js/commit/137784b098703a35f8912690f8cac4971e70e2e7 :+1:

bobylito:
:+1: merging!

Issue #46, opened 2015-04-12
Every actions that may change the number of pages should reset the current page to the first


bobylito:
When we do a refinement right now, it doesn't reset the current, leading to inconsistencies.

PR #52, opened 2015-04-14
Set page to 0 when mutating the search parameters

bobylito:
Fix for #46


Jerskouille:
So, how do we handle the `mutateMe` object creation on this one ? If we chose not to, :+1: .

bobylito:
As discussed yesterday, we want to avoid too much object creations. I'll move the modifications directly in the SearchParameters.

Jerskouille:
:1: !

PR #56, opened 2015-04-20
Fix #53 : handle numeric filters for attributes declared as disjunctive facets

### LINKED ISSUE #53: What about numeric facettings ?
Right now, if we want a slider, we have an issue : a `disjunctiveFacet` doesn't ignore the `numericRefinements` on itself, which could make sense in some cases, but kills the whole idea of using facetting for this since the facet stats which come back take into account the newly created `numericRefinements`. How do we handle it ?
This question might also bring the same question with `facetsExcludes`, but I'm pretty sure that I we exclude a value from a disjunctive facet, we really don't want it at all.

bobylito:
The ability to declare a numeric attribute as a disjunctive facet makes it easy to create sliders in instant search implementations.

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -1,5 +1,6 @@
 "use strict";
 var keys = require( "lodash/object/keys" );
+var intersection = require( "lodash/array/intersection" );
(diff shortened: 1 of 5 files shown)


Jerskouille:
:+1:

PR #61, opened 2015-04-24
Fix #60 : counts for refined disjunctive facets not correct

### LINKED ISSUE #60: Count of facets not retrieved
Test case on the sample ( http://algolia.github.io/algoliasearch-helper-js/examples/instantsearch+helper.html ):
 - the value "marshall" for the facet manufacturer is not displayed 
 - search for "marshall" in query input
 - the value "marshall" is displayed with the count equals to 5
 - check the "marshall" facet value in manufacturer
 - value for this facet is still *5*
 - remove marshall from the query input
 - value is *0* but should be *5*

bobylito:


redox:
OK, this one is not really a bug: it's expected. Retrieving all the facets from the main (hits) query might fix some cases, not all of them; that being said it's good enough to merge.

PR #67, opened 2015-05-07
add setTypoTolerance to SearchParameters

maxiloc:


src/SearchParameters/index.js
```diff
@@ -182,6 +182,20 @@ SearchParameters.prototype = {
...
+   * typoTolerance setter
+   * Set the value of typoTolerance
+   * @method
+   * @param {boolean} b boolean new value of typoTolerance
+   * @return {SearchParameters}
+   */
+  setTypoTolerance : function setTypoTolerance( b ) {
+    return this.mutateMe( function( m ) {
+      m.typoTolerance = b;
+      m.page = 0;
```
bobylito: Good catch to reset the current page!

src/SearchParameters/index.js
```diff
@@ -182,6 +182,20 @@ SearchParameters.prototype = {
...
+
+  /**
+   * typoTolerance setter
+   * Set the value of typoTolerance
+   * @method
+   * @param {boolean} b boolean new value of typoTolerance
+   * @return {SearchParameters}
+   */
+  setTypoTolerance : function setTypoTolerance( b ) {
+    return this.mutateMe( function( m ) {
```
bobylito: It's not a boolean. it's a string with the values "true", "false" , "min" or "strict" http://algolia.github.io/algoliasearch-helper-js/docs/SearchParameters.html#typoTolerance https://www.algolia.com/doc/rest_api#typoTolerance

redox: But from a JS POV it can be a boolean OR a string: ```true || false || "min" || "strict"```

bobylito:
:+1: Awesome! Thanks for your first contribution!

PR #74, opened 2015-05-12
Fix/73

bobylito:
@pixelastic @redox Please :D

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -259,10 +259,10 @@ AlgoliaSearchHelper.prototype.setState = function( newState ) {
  *  } );
  *  window.onpopstate = function( event ){
  *    // This is naive though as you should check if the state is really defined etc.
- *    helper.overrideState( event.state ).search();
+ *    helper.overrideStateWithoutTriggeringChangeEvent( event.state ).search();
  *  }
  */
-AlgoliaSearchHelper.prototype.overrideState = function( newState ) {
+AlgoliaSearchHelper.prototype.overrideStateWithoutTriggeringChangeEvent = function( newState ) {
   this.state = new SearchParameters( newState );
   return this;
 };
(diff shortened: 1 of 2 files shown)


src/algoliasearch.helper.js
```diff
@@ -247,6 +247,27 @@ AlgoliaSearchHelper.prototype.setState = function( newState ) {
...
+ *  helper.on( "change", function( state ){
+ *    // In this function you might want to find a way to store the state in the url/history
+ *    updateYourURL( state );
+ *  } );
+ *  window.onpopstate = function( event ){
+ *    // This is naive though as you should check if the state is really defined etc.
+ *    helper.overrideState( event.state ).search();
+ *  }
+ */
+AlgoliaSearchHelper.prototype.overrideState = function( newState ) {
```
pixelastic: We should name it with something less ambiguous. It is not clear from the method signature what the difference between `setState` and `overrideState` will be.

I haven't found a better name yet, but here are some ideas:
- `setStateWithoutTriggeringChangeEvent`
- `setStateSilent`
- `setState(newState, { 'silent': true })`
- `_setState`

Jerskouille:
Okay, about the URL that you don't want to handle @bobylito , why not have this function actually be something like `loadState` so that we could have : 

```js
var prefix = '#state=';

index.on('change', function () {
  window.location.hash = prefix + encodeURIComponent(index.serializedState());
});

if (window.location.hash && window.location.hash.length > prefix.length) {
    index.loadState(decodeURIComponent(window.location.hash.substring(prefix.length)));
}
```

Like this, it would still make sense to have this even on node's side, while not actually touching `window`.

bobylito:
@Jerskouille We don't need another function to do the initial setup of the state, because the constructor of the helper takes as a third parameter the options, which is a partial object of the SearchParameters.

The override function is here to give you the ability to update the state in the same fashion without having to create a new instance of the helper. We want this because this will be tricky to propagate in an already setup application.

Here is an example of how I see the use of this  feature : 
https://github.com/algolia/algolia-ui-kit/blob/feature/url/src/index.js#L47-L66

vvo:
Aside from the method name which is very react-y (in the sense of descriptive), LGTM! :-)

bobylito:
Cool! merging! Thanks all for the comments!

PR #75, opened 2015-05-12
Fix/72

bobylito:
@redox @pixelastic @Jerskouille @maxiloc @vvo Review please? :)

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -365,9 +365,16 @@ SearchParameters.prototype = {
    * @return {SearchParameters}
    */
   addFacetRefinement : function addFacetRefinement( facet, value ) {
+    if( this.isRefined( facet, value ) ) {
+      return this;
+    }
+
     return this.mutateMe( function( m ) {
       m.page = 0;
-      m.facetsRefinements[ facet ] = value;
+      if( !m.facetsRefinements[ facet ] ) {
+        m.facetsRefinements[ facet ] = [];
+      }
+      m.facetsRefinements[ facet ].push( value );
     } );
   },
   /**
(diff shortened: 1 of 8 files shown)


src/SearchParameters/index.js
```diff
@@ -365,9 +365,16 @@ SearchParameters.prototype = {
    * @return {SearchParameters}
    */
   addFacetRefinement : function addFacetRefinement( facet, value ) {
+    if( this.isRefined( facet, value ) ) {
+      return this;
```
pixelastic: Shouldn't it return a `this.mutateMe(noop)` instead ? So the return of `addFacetRefinement` is consistenly a new object ?

bobylito: In the case isRefined( facet, value ) returns true, it would mean that there would be no changes, so might as well avoid creating a new instance (there are thousands of case we create useless new objects, true :D, it's just continuous improvement here)

src/SearchResults/index.js
```diff
@@ -133,7 +133,7 @@ var SearchResults = function( state, algoliaResponse ) {
     forEach( result.facets, function( facetResults, dfacet ) {
       var position = disjunctiveFacetsIndices[ dfacet ];
 
-      var dataFromMainRequest = mainSubResponse.facets[ dfacet ];
+      var dataFromMainRequest = mainSubResponse.facets && mainSubResponse.facets[ dfacet ] || {};
```
Jerskouille: Parenthesis would be nice here.
Also, isn't `mainSubResponse.facets[ dfacet ]` is an array, so shouldn't it be `[]`?

Jerskouille:
Except that, LGTM. :+1:

redox:
:+1: LGTM

src/SearchResults/index.js
bobylito: In the response we get from algolia, the values of a single facet is an object of the form object<attributeValue : string, count: number>  https://www.algolia.com/doc#FacetNavigation . Thus it has to be consistent with the object merge we do two lines after https://github.com/algolia/algoliasearch-helper-js/blob/fix/72/src/SearchResults/index.js#L139

src/SearchParameters/index.js
```diff
@@ -403,15 +410,29 @@ SearchParameters.prototype = {
...
+  removeFacetRefinement : function removeFacetRefinement( facet, value ) {
     return this.mutateMe( function( m ) {
       m.page = 0;
-      m._clearFacetRefinements( facet );
+      if( value ) {
+        var idx = m.facetsRefinements[ facet ].indexOf( value );
+        if( idx > -1 ) {
+          m.facetsRefinements[ facet ].splice( idx, 1 );
+          if( m.facetsRefinements[ facet ].length === 0 ) {
+            delete m.facetsRefinements[ facet ];
```
vvo: why do we need to delete the array?

bobylito: actually we remove the key from the object so when iterating with a `for in` we don't need to check for the content.

vvo:
LGTM

bobylito:
Thanks for the comments :) Merging!

PR #81, opened 2015-05-15
Expose the exhaustiveFacetsCount flags.

redox:


bobylito:
Thanks @redox. Is it good for peer review? :)

Jerskouille:
Ok, made the test pass. And the timeouts was definitely not used yet, since we werre calling `assignFacetTimeout` with parameters in the wrong order. (I think it was my mistake)

bobylito:
LGTM thanks guys :)

PR #83, opened 2015-05-19
Fix #82 : Fix facet count on records with facetted attribute with mul…

### LINKED ISSUE #82: Inconsistent facet count when records have multiple values for a single facet

bobylito:
…tiple values


bobylito:
Manual merge :)

PR #88, opened 2015-06-01
Fix #69 : Ability to know if a facet is refined, whatever the value

### LINKED ISSUE #69: Ability to know if a facet is refined, whatever the value
Something like `helper.hasRefinements(facet) -> boolean`

bobylito:


redox:
LGTM but while reviewing it I saw some convenience method names that could be improved:

```js
hasRefinements: function(facet) { return isFacetRefined(facet); };
```

```js
// alias isExcludeRefined(facet, values) to isExcluded(facet, values)
```

What do you think?

vvo:
LGTM

bobylito:
@redox like the idea, is it ok to have this interface only for the helper. Or should it be as explicit in the SearchParameters as well?

PR #93, opened 2015-06-03
Fix #51 : conjunctive facets with multiple filtering

### LINKED ISSUE #51: Conjunctive facets aren't stored in a array
It works fine except when facetting an array attribute.

bobylito:
This PR is not a new feature. It tests that the feature is actually there. The tests implemented cover the different aspects required for the feature : SearchParameters can use many refinements for a single conjunctive facet, that those refinements lead to correct filtering in the query for the js client and finally that we get the expected results from Algolia engine.

It also provides the framework for integration testing with Algolia engine.

Special thanks to @vvo for the help on the integration tests ;)

### DIFF
--- a/test/spec/helper.facetFilters.js
+++ b/test/spec/helper.facetFilters.js
@@ -1,0 +1,29 @@
+//Make sure we do facet filters correctly before sending them to algolia
+
+"use strict";
+
+var test = require( "tape" );
+var algoliasearchHelper = require( "../../index" );
+
+test( "The filters should contain the different filters for a single conjunctive facet with multiple refinements",
+      function( t ) {
+  var facetName = "myFacet";
+  var helper = algoliasearchHelper( "", "", {
+    facets : [ facetName ]
+  } );
+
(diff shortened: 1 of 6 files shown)


test/spec/helper.facetFilters.js
```diff
@@ -0,0 +1,29 @@
...
+  helper.addRefine( facetName, "value2" );
+  t.deepEqual( helper._getFacetFilters(),
+               [ facetName + ":value1", facetName + ":value2" ],
+               "Two values : two filters" );
+  helper.addRefine( facetName, "value1" );
+  t.deepEqual( helper._getFacetFilters(),
+               [ facetName + ":value1", facetName + ":value2"],
+               "Add multiple times the same parameter : only two parameters" );
+
+  t.end();
```
redox: Could you also test the `toggle` and `remove` functions?

bobylito: Great idea! Will also update the unit tests.

Jerskouille:
LGTM ! :+1:

PR #94, opened 2015-06-03
Fix #92 parameters that can be set in the dashboard shouldn't have defaults

### LINKED ISSUE #92: Do not set default values that are computed in the API server
For example hitsPerPage and maxValuesPerFacet

bobylito:
@maxiloc @vvo ;)


vvo:
:+1: :shipit:

bobylito:
Thanks!

PR #95, opened 2015-06-04
 FIX #70 Ability to get all refined values : helper.getRefinements()

### LINKED ISSUE #70: Ability to get all refined values
Something like `helper.getRefinements(facet) -> ["value1", "value2", ...]`

bobylito:


src/SearchParameters/index.js
```diff
@@ -312,6 +312,30 @@ SearchParameters.prototype = {
     } );
   },
   /**
+   * Get the list of conjunctive refinements for a single facet
+   * @param {string} facetName
+   * @return {string[]} list of refinements
+   */
+  getConjunctiveRefinements : function( facetName ) {
+    return this.facetsRefinements[ facetName ];
```
redox: I would append `|| []` to ease its usage. What do you think? (same for all others)

src/algoliasearch.helper.js
```diff
@@ -343,6 +343,61 @@ AlgoliaSearchHelper.prototype.getCurrentPage = function() {
...
+        operator : operator,
+        type : "numeric"
+      } );
+    } );
+  }
+
+  return refinements;
+};
+
 ///////////// PRIVATE
```
redox: Oh OK, so here it returns an array even if it doesn't exist. Could make sense to only have it here!

src/algoliasearch.helper.js
```diff
@@ -343,6 +343,61 @@ AlgoliaSearchHelper.prototype.getCurrentPage = function() {
   return this.state.page;
 };
 
+/**
+ * Get the list of refinements for a give facet.
```
redox: `s/facet/attribute/` I would say, because of the numericFilters and the `_tags` very soon.

bobylito:
Could make sense overall. Making the changes!

src/algoliasearch.helper.js
bobylito: Where do you put your comment?! I can't outdate it :)

bobylito:
Seems like we're all done here. Merging!

PR #96, opened 2015-06-04
FIX #64 sum up the processing time of all the subqueries and the main

### LINKED ISSUE #64: The `processingTimeMS` should be summed
When we're dealing with disjunctive refinements, it generates multiple queries -> we need to sum the processingTimeMS of each answer to compute the final one.

bobylito:


src/SearchResults/index.js
```diff
@@ -75,7 +75,13 @@ var SearchResults = function( state, algoliaResponse ) {
    * processing time of the main query
    * @member {number}
    */
-  this.processingTimeMS = mainSubResponse.processingTimeMS;
+  this.processingTimeMS = ( function( responses ) {
```
vvo: seems like a good case to use a reduce?

vvo: => good opportunity

pixelastic: Or using lodash `_.sum(responses, 'processingTimeMS')`

bobylito: agreed thanks guys!

bobylito:
I guess we're done here :)

PR #97, opened 2015-06-05
FIX #29 Avoid callback for outdated queries

### LINKED ISSUE #29: Avoid callback for outdated queries
Depending on the network, it may happen that a request sent before another is received after. This should not happen, and shouldn't be left to the responsibility of the user.

bobylito:


redox:
LGTM :+1:

bobylito:
Thanks for the PR!

PR #99, opened 2015-06-08
Support for tags

bobylito:
This pull request makes the helper handles the tags in a similar fashion than the facets, the numerics or the excludes.


src/SearchParameters/index.js
```diff
@@ -443,6 +454,22 @@ SearchParameters.prototype = {
     } );
   },
   /**
+   * addTagRefinement adds a tag to the list used to filter the results
+   * @param {string} tag tag to be added
+   * @return {SearchParameters}
+   */
+  addTagRefinement : function addTagRefinement( tag ) {
+    if( this.tagFilters ) {
+      return this.setQueryParameter( "tagFilters", undefined )
```
redox: Re-thinking about that, I would rather go for an error/exception because it doesn't make sense to "loose" previous filters. Such "hidden" behavior are never understood correctly. What do you think?

bobylito: Yeah that sure sounds confusing, if the user didn't make that on purpose. But "if" it's done on purpose, it is just not possible to go from one to the other.

Imagine that in an UI, the developper wants to propose a simple checkbox list for handling tags and an advanced in which the user input the string himself.

redox: If it's on purpose he could do `helper.clearTagRefinements().addTagRefinement("foo").search()` or something like that. We need to ensure he can do it (with `clearTagRefinements()` or `setQueryParameters('tagFilters', undefined)` but I would not do it automagically.

bobylito:
On hold until I finish the error handling with the correct error messages.

redox:
LGTM :+1:

bobylito:
Yeah! :) Merging!

PR #100, opened 2015-06-08
Fix #62 : update rule when distinct can be applied

### LINKED ISSUE #62: Fix the case where the distinct is applicable
@redox said : 

Beware, you can use distinct=true if there are no facet filters, but some tag filters or some numeric filters or a query.

So probably something like:

```
if (!query && facetFilters.length === 0 && numericFilters.length === 0 && tagFilters.length === 0) {
  distinct = false;
}
```

bobylito:


redox:
LGTM :+1:

vvo:
mario

bobylito:
Yoohoo!

PR #103, opened 2015-06-10
FIX #25 Advanced clear refinement

### LINKED ISSUE #25: Clear refinements matching a specific criteria

bobylito:
Let the user use a function to clear specific refinements.

Will go through all the refinements, and call the user function with the parameters : 
 - value : for numeric the value and the operator, otherwise just the facet value
 - attribute : the attribute the refinement is applied on
 - type : the type of refinement

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -4,6 +4,7 @@ var intersection = require( "lodash/array/intersection" );
 var forEach = require( "lodash/collection/forEach" );
 var reduce = require( "lodash/collection/reduce" );
 var filter = require( "lodash/collection/filter" );
+var pick = require( "lodash/object/pick" );
 var isEmpty = require( "lodash/lang/isEmpty" );
 var isUndefined = require( "lodash/lang/isUndefined" );
 var isString = require( "lodash/lang/isString" );
@@ -250,12 +251,12 @@ SearchParameters.prototype = {
    * @return {SearchParameters}
    */
   clearRefinements : function clearRefinements( name ) {
-    return this.mutateMe( function( m ) {
-      m.page = 0;
-      m._clearNumericRefinements( name );
-      m._clearFacetRefinements( name );
-      m._clearExcludeRefinements( name );
-      m._clearDisjunctiveFacetRefinements( name );
+    return this.setQueryParameters( {
+      page : 0,
+      numericRefinements : this._clearNumericRefinements( name ),
+      facetsRefinements : this._clearFacetRefinements( name ),
+      facetsExcludes : this._clearExcludeRefinements( name ),
+      disjunctiveFacetsRefinements : this._clearDisjunctiveFacetRefinements( name )
     } );
   },
   /**
@@ -436,12 +437,12 @@ SearchParameters.prototype = {
    */
   _clearNumericRefinements : function _clearNumericRefinements( attribute ) {
     if ( isUndefined( attribute ) ) {
-      this.numericRefinements = {};
+      return {};
     }
     else if ( isString( attribute ) ) {
-      if ( !isUndefined( this.numericRefinements[ attribute ] ) ) {
-        delete this.numericRefinements[ attribute ];
-      }
+      return pick( this.numericRefinements, function( value, key ) {
+        return attribute !== key;
+      } );
     }
   },
   /**
@@ -533,7 +534,7 @@ SearchParameters.prototype = {
         }
       }
       else {
-        m._clearFacetRefinements( facet );
+        m.facetsRefinements = m._clearFacetRefinements( facet );
       }
     } );
   },
@@ -606,12 +607,12 @@ SearchParameters.prototype = {
    */
   _clearFacetRefinements : function _clearFacetRefinements( facet ) {
     if ( isUndefined( facet ) ) {
-      this.facetsRefinements = {};
+      return {};
     }
     else if ( isString( facet ) ) {
-      if ( !isUndefined( this.facetsRefinements[ facet ] ) ) {
-        delete this.facetsRefinements[ facet ];
-      }
+      return pick( this.facetsRefinements, function( value, key ) {
+        return key !== facet;
+      } );
     }
   },
   /**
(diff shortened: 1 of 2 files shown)


bobylito:
Ping @redox @vvo @Jerskouille @pixelastic, review welcome :)

src/SearchParameters/index.js
```diff
@@ -429,19 +434,30 @@ SearchParameters.prototype = {
...
   _clearNumericRefinements : function _clearNumericRefinements( attribute ) {
     if ( isUndefined( attribute ) ) {
-      this.numericRefinements = {};
+      return {};
     }
     else if ( isString( attribute ) ) {
-      if ( !isUndefined( this.numericRefinements[ attribute ] ) ) {
-        delete this.numericRefinements[ attribute ];
-      }
+      return omit( this.numericRefinements, function( value, key ) {
```
pixelastic: I think this can be simplified to `_.omit(this.numericRefinements, attribute)`. I think lodash exposes this kind of sugar.

test/spec/helper.clears.js
```diff
@@ -68,6 +71,58 @@ test( "Clearing the same field from multiple elements should remove it everywher
   t.end();
 } );
 
+test( "Clear with a function : neutral and abosorbing ops", function( t ) {
```
pixelastic: aborsorbing? ;)

test/spec/helper.clears.js
```diff
@@ -68,6 +71,58 @@ test( "Clearing the same field from multiple elements should remove it everywher
...
+  helper.clearRefinements( function( value, key, type ) {
+    return false;
+  } );
+
+  t.deepEqual( helper.state.numericRefinements, state0.numericRefinements, "Neutral op : numeric ref should be equal" );
+  t.deepEqual( helper.state.facetsRefinements, state0.facetsRefinements, "Neutral op : conj ref should be equal" );
+  t.deepEqual( helper.state.facetsExcludes, state0.facetsExcludes, "Neutral op : exclude ref should be equal" );
+  t.deepEqual( helper.state.disjunctiveFacetsRefinements, state0.disjunctiveFacetsRefinements, "Neutral op : disj ref should be equal" );
+
+  helper.clearRefinements( function( value, key, type ) {
```
pixelastic: I would split those two cases in two tests : neutral and remove all.

test/spec/helper.clears.js
```diff
@@ -68,6 +71,58 @@ test( "Clearing the same field from multiple elements should remove it everywher
...
+  t.assert( isEmpty( helper.state.numericRefinements ), "remove all numericRefinements" );
+  t.assert( isEmpty( helper.state.facetsRefinements ), "remove all facetsRefinements" );
+  t.assert( isEmpty( helper.state.facetsExcludes ), "remove all facetsExcludes" );
+  t.assert( isEmpty( helper.state.disjunctiveFacetsRefinements ), "remove all disjunctiveFacetsRefinements" );
+
+  t.end();
+} );
+
+test( "Clear with a function : filtering", function( t ) {
+  var helper = init();
```
pixelastic: I don't really like tests that depends on a boilerplate defined somewhere else, used for multiple tests. It does not make tests self explanatory.

I don't mind writing something like:

```
helper.toggleRefine( "keepFacet", "0" )
    .toggleRefine( "dropFacet", "0" )
    .toggleRefine( "keepDisjunctiveFacet", "0" )
    [...]

helper.clearRefinements( function( value, key, type ) {
    return key.indexOf( "drop" ) === 0;
  } );
```

More verbose in each test, but more explicit as the content we test is only used for that test in the end. WDYT?

src/SearchParameters/index.js
```diff
@@ -599,57 +615,90 @@ SearchParameters.prototype = {
...
     else if ( isString( facet ) ) {
-      if ( !isUndefined( this.disjunctiveFacetsRefinements[ facet ] ) ) {
-        delete this.disjunctiveFacetsRefinements[ facet ];
-      }
+      return omit( this.disjunctiveFacetsRefinements, function( value, key ) {
+        return key === facet;
+      } );
+    }
+    else if ( isFunction( facet ) ) {
+      return reduce( this.disjunctiveFacetsRefinements, function( memo, values, key ) {
```
pixelastic: Maybe `_.compact(_.map())` would be easier than `_.reduce` ?

test/spec/helper.clears.js
bobylito: Yeah that's not a good and easy to understand. It's like in math, where 0 is absorbant (almost sure it's a wrong translation) for the multiplication...

src/SearchParameters/index.js
bobylito: I'm not sure.

src/SearchParameters/index.js
bobylito: Yep! Will fix it!

test/spec/helper.clears.js
bobylito: why not...

test/spec/helper.clears.js
bobylito: Not sure, we might not end up with just a bunch of copy paste of the init...

test/spec/helper.clears.js
bobylito: I changed that!

vvo:
LGTM and understood the underlying feature: nice job!

vvo:
To ease reviewing you could also add an example usage of the new feature in the PR description.
Would realy help to get the context easily. (next time?)

test/spec/helper.clears.js
pixelastic: The idea is to only init in the test the parts we actually want to test. I don't like copy-paste either, but I also don't like having to deeply know what's in an init/fixture to understand a test. It's an open question and I don't have a definitive answer to that.

vvo: When I read the test I definitely thought the "init()" part was maybe not the best name. It's a fixture let's call it this way?

vvo: Otherwise once I saw it and once I saw the fixture was tested then it's all good?

pixelastic: Yep, renaming it to fixture could make it better. I guess that's the fact that we have one fixture to test multiple cases that bugs me. It makes each test case more complex than it should be.

But here we're discussing meta, that should not block this PR. :+1:

bobylito: If it's a fixture, then it is :) (shame on me I'm not familiar with this name :$)

bobylito:
Yes you're right @vvo , will add more to the description next time :)

bobylito:
Thanks @vvo @pixelastic for taking the time to review. Seems like everyone is happy, MERGIN'

PR #104, opened 2015-06-10
s/integation/integration/

pixelastic:
Just renamed a directory that buggued me ;)


bobylito:
Thanks. The benefits and pitfall of autocomplete :)

PR #105, opened 2015-06-12
Make the core model/state of the search (SearchParameter) immutable (for real)

bobylito:
The point of this pull request is to provide real immutability for the state of the search that is encapsulated into the helper (`helper.state`)

In practice, this will help to easily test that the previous state is different from the current by comparing references. It will also help test properties for the future, you can now easily modify the model without having to revert the changes.

```javascript
var state0 = SearchParameters.make( {
  disjunctiveFacets : [ "facet" ]
} );

var state1 = state0.addDisjunctiveRefinement( "facet", "value" );

state0 !== state1 //true
deepEqual( state0, state1 ); // false

//actually

state0.disjunctiveRefinements !== state1.disjunctiveRefinements;
```


src/SearchParameters/RefinementList.js
```diff
@@ -0,0 +1,130 @@
...
+    if( lib.isRefined( refinementList, attribute, value ) ) {
+      return lib.removeRefinement( refinementList, attribute, value );
+    }
+
+    return lib.addRefinement( refinementList, attribute, value );
+  },
+  /**
+   * Clear all or parts of a RefinementList. Depending on the arguments, three
+   * behaviors can happen :
+   *  - if not attribute is provided : clears the whole list
```
pixelastic: s/not/no/

src/SearchParameters/index.js
```diff
@@ -331,9 +354,11 @@ SearchParameters.prototype = {
    * @return {SearchParameters}
    */
   setHitsPerPage : function setHitsPerPage( n ) {
-    return this.mutateMe( function( m ) {
-      m.hitsPerPage = n;
-      m.page = 0;
+    if( this.HitsPerPage === n ) return this;
```
pixelastic: Shouldn't this be `hitsPerPage` (lowercase) ?

src/SearchParameters/index.js
```diff
@@ -344,9 +369,11 @@ SearchParameters.prototype = {
    * @return {SearchParameters}
    */
   setTypoTolerance : function setTypoTolerance( s ) {
```
pixelastic: Can't we use a more explicit argument name, like in the in `setPage` or `setQuery`, like `typoTolerance` ?

pixelastic:
LGTM :+1:

vvo:
✅ reviewed

bobylito:
Thanks @vvo @pixelastic

PR #109, opened 2015-06-16
Fix #107 : fix rules for distinct

### LINKED ISSUE #107: Distinct should default to false only when the user tries to set it
The value for distinct should only be set in the case the user sets it because he might have set it in its dashboard. @pixelastic reported this one.

bobylito:
Rules work as follows :
 - if no distinct and no filters are applied : force distinct to false
 - if no distinct and some filters : do not set the distinct
 - if distinct set to true and no filters : force distinct to false
 - if distinct set to true and some filters : let the distinct as true
 - if distinct set to false (with or without filters) : let the distinct set to false


bobylito:
@pixelastic @redox , cool for you?

pixelastic:
> if distinct set to true and no filters : force distinct to false
> if distinct set to true and no filters : let the distinct as true

Which one?

bobylito:
erf fixed.

src/algoliasearch.helper.js
```diff
@@ -537,10 +537,13 @@ AlgoliaSearchHelper.prototype._getHitsSearchParams = function() {
   var additionalParams = {
     facets : facets,
     tagFilters : tagFilters,
-    distinct : this.state.distinct || false
+    distinct : this.state.distinct
   };
 
-  if( !this.state.query && facetFilters.length === 0 && numericFilters.length === 0 && tagFilters.length === 0 ) {
+  if( !this.state.query &&
```
pixelastic: We might want to extract this check into a `isQueryEmpty()` method, as we're using it in several places. This will also make the intent more explicit.

test/spec/helper.distinct.js
```diff
@@ -0,0 +1,127 @@
...
+var test = require( "tape" );
+
+var algoliasearchHelper = require( "../../index.js" );
+
+test( "Distinct not set", function( t ) {
+  var helper = algoliasearchHelper( null, null, {} );
+  var state0 = helper.state;
+
+  var disjunctiveFacetSearchParam = helper._getDisjunctiveFacetSearchParams();
+  t.equal( disjunctiveFacetSearchParam.distinct, false, "[disjunctive] distinct should be undefined" );
```
pixelastic: Aren't you testing that it's false here?

pixelastic:
Except for this two minor comments, LGTM. Tell me when it's available on jsdelivr.

src/algoliasearch.helper.js
bobylito: +1

test/spec/helper.distinct.js
bobylito: Arf

bobylito:
Mergin' @pixelastic will tell you when it's out ;)

PR #114, opened 2015-06-23
Fix #112 : do not set undefined to distinct in the request.

### LINKED ISSUE #112: distinct=undefined is sent to the API if distinct is not set
distinct=undefined is sent to the API if distinct is not set, generating a 400 error.

bobylito:


vvo:
it seems you could get rid of the refactor setup in this PR by starting from develop instead, or closing the PR and reopening another on for develop.

vvo:
Also we need a new test for the bugfix I guess

bobylito:
Argh :/ bad branch...

PR #118, opened 2015-06-23
FIX #110 : provide a way to test if an attribute is numerically defined

### LINKED ISSUE #110: isRefined is not checking if the attribute has a numeric refinement
`isRefined` is checking if it a facet or disjunctive facet has a refinement; we should also check it it's an attribute with a numerical refinement.

bobylito:
hasRefinements now test if a given attribute contains a numeric filter too.

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -5,6 +5,7 @@ var extend = require( "./functions/extend" );
 var util = require( "util" );
 var events = require( "events" );
 var forEach = require( "lodash/collection/forEach" );
+var isEmpty = require( "lodash/lang/isEmpty" );
 var bind = require( "lodash/function/bind" );
 
 /**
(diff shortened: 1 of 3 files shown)


bobylito:
ping @redox :)

src/algoliasearch.helper.js
```diff
@@ -346,12 +347,12 @@ AlgoliaSearchHelper.prototype.isRefined = function( facet, value ) {
...
- * @param {string} facet the facet attribute name
- * @return {boolean} true if the facet is facetted by at least one value
+ * Check if the attribute has any numeric, disjunctive or conjunctive refinements
+ * @param {string} attribute the name of the attribute
+ * @return {boolean} true if the attribute is filtered by at least one value
  */
-AlgoliaSearchHelper.prototype.hasRefinements = function( facet ) {
-  return this.isRefined( facet );
+AlgoliaSearchHelper.prototype.hasRefinements = function( attribute ) {
+  return !isEmpty( this.state.getNumericRefinements( attribute ) ) || this.isRefined( attribute );
```
pixelastic: From an end-user POV, I don't really get the difference between `isRefined` and `hasRefinements`. Is `isRefined` a lower level method?

If you had to split the two boolean checks (`!isEmpty( this.state.getNumericRefinements( attribute ) )` and `this.isRefined( attribute )` in variables, how would you call them? This could make the code easier to grasp.

redox: Yes, I agree.

bobylito: Well for the two comments : 
 - we can't provide an API that fits the structure of both the facets (attribute, value) and the numerics (attribute, operator, value), the same way we can't provide that for tags (value). So we need those two methods, one that is specific to facets (isRefined) and one that is specific to attribute based refinements (hasRefinements)
 - Ok I will clean that up so it's easier to get what's going when reading the code.

bobylito:
Mergin', let's discuss further changes of the API for the next major release :)

redox:
:+1: @bobylito

PR #119, opened 2015-06-24
Fix #106 Convert facet value to string

### LINKED ISSUE #106: Always convert facet value to string in addRefinement
If you add a refinement with a integer value, you won't be able to know if the value is refined in the on('result') because the response returns a string value.

bobylito:
Use case : 
 - user input a number as a facet filter
 - algolia returns string for facet values
 - user can't loop over the values to check if they are refined because on one side the value is a string, and on the other the value is a number, making the comparison false.

In this fix, the lib will always convert any input value for the filters to strings. This will apply to conjunctive, disjunctive and exclude refinements.


bobylito:
@Nagriar this should be of interest to you ;)

PR #129, opened 2015-07-01
test: first pass at full CI browser/node/iojs testing

vvo:
This pull request activates travis ci testing on master/develop and
pull requests.

In details:
+ CI testing on all currently supported browsers
+ CI testing on node 0.10, 0.12, iojs
+ `npm test` command that will: test phantom, test node, lint

Package upgrades:
- zuul: gzip assets + better error reporting (defunctzombie/zuul#220,
defunctzombie/zuul#221)
- browserify, eslint, sinon, tape, uglify-js

Various:
+ put `latest` for algoliasearch dev dependency, enforce being always on
the latest version when testing
+ some docs around contributing
+ BADGES!

2015-07-02-002043_960x408_scrot
2015-07-02-002016_273x35_scrot


vvo:
@bobylito I had to .skip() integration tests because they are failing. Since they are also failing on `master`, I assume they were already broken.

If you can have a look that would be cool.

bobylito:
Thanks for this awesome PR.

I will have a look. Note that these tests sometimes fail, and sometimes run
correctly... I hope this is what you encountered. I will make them more
reliable anyway :)

vvo:
merge #131 then #129 please :)

bobylito:
#131 has been closed without merged... Should I reopen it?

CONTRIBUTING.md
```diff
@@ -0,0 +1,12 @@
+How to contribute?
+
+Basic flow:
+- fork this repository
+- setup the project: `npm install`
+- start dev mode `npm run dev` and open http://localhost:8090/__zuul
```
bobylito: Well maybe we should remove this line or add a step with the following statement : "make your contribution"?

vvo:
No I will simplify the linting process fir both us and CI/tests you can merge this one before no problem

README.md
```diff
@@ -2,11 +2,25 @@
 
 # algoliasearch-helper-js
 
-This module is the companion of the algoliasearch-client-js. It helps you keep
+This module is the companion of the [algoliasearch-client-js](https://github.com/algolia/algoliasearch-client-js). It helps you keep
 track of the search parameters and provides a higher level API.
 
-The helper is built on top of algoliasearch-client-js and this version is 
-specifically made to work with the newest V3 versions of it.
+This is the plugin you will need to easily build a good search UX.
```
bobylito: Good thing that you add an explanation here. On the vocabulary side, I would say that this is more of a wrapper than a plugin. Do you think that plugin will be more easily understood?

scripts/test-node.sh
```diff
@@ -0,0 +1,31 @@
...
+  nvm use 0.10
+  node test/run.js
+
+  echo "Node test iojs"
+  nvm use iojs
+  node test/run.js
+fi
+
+# switch back to node.js 0.12
+nvm use 0.12
```
bobylito: is there a way to go back to the user version?

test/spec/helper.pages.js
```diff
@@ -34,11 +34,16 @@ test( "previousPage should decrement the current page by one", function( t ) {
...
 test( "pages should be reset if the mutation might change the number of pages", function( t ) {
+  var bindAll = require( "lodash/function/bindAll" );
+  var partial = require( "lodash/function/partial" );
+
   var helper = algoliasearchHelper( "", "", {
     facets : [ "facet1", "f2" ],
     disjunctiveFacets : [ "f1" ]
   } );
 
+  bindAll(helper);
```
bobylito: That's a bit invasive. Aren't there cases, where this could make something go wrong?

bobylito:
OK did a review. This is very good material :) I like the refactoring of the scripts especially. I guess this will make our projects similar now :)

@vvo can you answer my remarks, and then we'll be good to go :)

CONTRIBUTING.md
vvo: what about two lines below where I added:

```
code, add a [test](./test/) for every fix or new feature
```
?

README.md
vvo: We can say "This is the library you need" so that we do not have to choose on whether this is a plugin or a wrapper (and yet this is maybe more a wrapper yes, but also a plugin :racehorse:)

test/spec/helper.pages.js
vvo: not that I can think of, also this may be unecessary since I partial() on methods thus the context should not be modified but not sure

vvo:
Good, will fix them soon :)

README.md
bobylito: Good for me! :)

CONTRIBUTING.md
bobylito: Sounds great!

scripts/test-node.sh
vvo: yes, very good idea I found how

vvo:
Ok @bobylito, fixes done!

PR #131, opened 2015-07-02
chore: code style change, lint with algolia-standard

### LINKED ISSUE #128: use standard formatter
To end issues about style and stop having to reformat every line of code written by different people, we could use the `standard` formatter: https://github.com/feross/standard

- have a look at the predefined [rules](https://github.com/feross/standard#rules)
- there is no configuration only a standard style builtin
- there is an automatic formatter and [editor plugins](https://github.com/feross/standard#editor-plugins)
- it's basically like [gofmt](https://golang.org/cmd/gofmt/) but in JavaScript
- it's based on eslint

I would be happy to switch all our JavaScript projects to standard format and start building things the fastest way instead of wrapping our mind and our code to the current project lead preferences.

So that you never find yourself contributing to an Algolia project and using your valuable time to align things, put a space there and there, remove a space there, fight with parenthesis, brackets, equals. 

A good number of projects are using standard, including the whole [npm codebase](https://github.com/npm/npm/blob/8c9fe9ce196c7f99b374ec5e45a1298e27b6d6e7/package.json#L175).

What do you think?

vvo:
Following #128 and various discussions we already had on code style, I
implemented [algolia/algolia-standard](https://github.com/algolia/standard) in the project.

The goal of this PR is to provide a solid ground for an easy linter
that we can use on all our projects.

The currently picked coding style is the one from Airbnb, with
modifications.

You can find the rules here: https://github.com/algolia/standard#rules

We can discuss the rules of course as long as we do not discuss them
every week!

Other projects:

- Original feross/standard:
After trying it I discovered some rules were not strict enough and
would lead to again style discussions. Has some sort of formatter but
not very good.

- JSCS:
I tried JSCS: good but only oriented on style, no rules on unused
variables or ways to prevent bugs like eslint does. Has some sort of
formatter, but not very good.

Ultimately eslint will have a good formatter, in the meantime we will
all switch to some new coding style, for good :-)?

fixes #128

Note: I submitted this PR to ci-tests branch since it was not yet merged, only to have a smaller diff with not all the ci-tests commits. Before merging both, merge this one first, then test/ci-tests.


vvo:
Also, you can then use https://github.com/algolia/standard#editor-plugins

vvo:
I am gonna change how we lint you can merge the first PR i will reopen on develop

PR #132, opened 2015-07-02
jsdelivr versionning

Jerskouille:
Since we're using Semver, I believe we should recommend to use v2 instead of v2.0.0.
jsDelivr's cache lasts 7 days (and we can rebuild it manually if needed), which means that we have some room to fix an issue introduced in a version, while being able to push an update on our clients websites seamlessly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ helper.search();
 
 Include this in your page :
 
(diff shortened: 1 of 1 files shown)


vvo:
I also made this change in #129

Jerskouille:
Ok, then I close this one.

PR #135, opened 2015-07-06
test: fix integration tests failures

vvo:
+ add debugging when it fails
+ close the underlying nodejs keep alive agent when needed (destroy)


bobylito:
Nice trick to display errors with promises! Thanks!

PR #140, opened 2015-07-06
Fix #130 : toggleRefine should throw an error if facet is not def

### LINKED ISSUE #130: should the helper throw on undeclared facet toggleRefine() usage?
Referencing this code: https://github.com/algolia/algoliasearch-helper-js/blob/f8d33b99337700f325438f7778c0c78537850dbd/src/algoliasearch.helper.js#L222-L228

Should we throw instead of just console.log()? A console.log can silently fail and will maybe slow down the debugging from the user point of view.

While a throw would break the application and thus would force the developer to debug, and find the error message in console.

bobylito:
Instead of a console message


vvo:
You can rebase on develop + we need a test here you can use https://github.com/substack/tape#tthrowsfn-expected-msg for this for example

bobylito:
ping @vvo :)

PR #144, opened 2015-07-08
style: run jsfmt on current code

vvo:
- upgrade eslint configuration to latest (some spacing were not consistent)
- created [algolia/jsfmt](https://github.com/algolia/jsfmt) that can rewrite a whole project according to our style
- add an `npm run jsfmt` script
- run `jsfmt --diff` as part of the linting, so that when code style is not conform, we error. Most of the time, fixing the ESLint errors should get you out of the way of jsfmt.
- If you just want to reformat at any point, do `npm run jsfmt`

Also

```sh
npm install @algolia/jsfmt -g
```

To use it anywhere in any project.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 'use strict';
+
 var AlgoliaSearchHelper = require('./src/algoliasearch.helper');
 
 var SearchParameters = require('./src/SearchParameters');
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "doc:watch": "onchange 'src/**/*.js' 'index.js' 'tutorials/*.md' '../minami/**/*.*' -- npm run doc",
     "doc:publish": "gh-pages-deploy",
     "test-ci": "scripts/test-ci.sh",
-    "test": "scripts/test.sh"
+    "test": "scripts/test.sh",
+    "jsfmt": "jsfmt"
   },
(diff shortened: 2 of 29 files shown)


vvo:
- upgrade eslint configuration to latest (some spacing were not consistent)
- created [algolia/jsfmt](https://github.com/algolia/jsfmt) that can rewrite a whole project according to our style
- add an `npm run jsfmt` script, currently not launched automatically anywhere

So currently the only issue with jsfmt preventing us to use it automatically in `npm test` for instance, is that it will rewrite:

```js
return 'something' +
           'some other thing';
```

to

```js
return 'something' +
  'some other thing';
```

There is no way currently to fix this behavior nor it is enforcable with ESLint.

But still `jsfmt` works pretty well and you can use it from time to time. Maybe we'll find a better way to automate it soon without ruining our workflow.

Right now it is provided as a one shot command that you can run when you want.

Also

```sh
npm install @algolia/jsfmt -g
```

To use it anywhere in any project.

vvo:
cc @bobylito @pixelastic

vvo:
@bobylito I updated the PR with latest changes in eslintconfig and jsfmt, we are good now :)

bobylito:
lgtm :+1:  thanks :)

Issue #137, opened 2015-07-06
API proposal: hierarchicalFacets


vvo:
I recently had to do a POC using hierarchical facets navigation. Something like:

```
- All products (10)
  - food (10)
    - fruits (10)
      - bananas (2)
      - kiwis (3)
      - oranges (5)
        - portugal (5)
```

In this example, we clicked on "food" > "fruits" > "oranges" > "portugal"

This is a not so easy task to do and one that was asked by a good number of clients lately.
We already talked about it in #24 and it has been done differently by multiple people.

To simplify this use case for our clients, I propose to add a new API.

## Acknowledgments

For hierarchical faceting to work well, we'll need the index to have certain attributes storing the hierarchical path.

Those attributes can have different names but their values must reflect the hierarchical path.

Example:

```json
{
  "name": "Portuguese oranges",
  "categories": {
    "level0": "food",
    "level1": "food > fruits",
    "level2": "food > fruits > oranges",
    "level3": "food > fruits > oranges > portugal"
  }
}
```

Why do we need to have the full path at every level? Because you may have `oranges` category in another `level1` category like `food > smoothies`.

Thus when we will ask for `level2 === oranges` we will not be able to know what are the oranges we want, oranges fruits or oranges smoothies?

### Simple usage

We should aim for this simple usage most of the time.

Index objects:
```json
{
  "name": "Portuguese oranges",
  "categories": {
    "level0": "food",
    "level1": "food > fruits",
    "level2": "food > fruits > oranges",
    "level3": "food > fruits > oranges > portugal"
  }
}
```

JS code:
```js
var helper = algoliasearchHelper(client, indexName, {
  hierarchicalFacets: ['categories']
});

helper.toggleRefine('categories', 'food > fruits > oranges > portugal');

helper.search();

helper.on('result', function(content) {
  console.log(content.hierarchicalFacets);
  // [{
  //   name: 'categories',
  //   isRefined: true,
  //   path: '',
  //   count: 10,
  //   data: [{
  //     name: 'food',
  //     path: 'food',
  //     isRefined: true,
  //     count: 10,
  //     data: [{
  //       name: 'fruits',
  //       path: 'food > fruits',
  //       isRefined: true,
  //       count: 10,
  //       data: [{
  //         name: 'bananas',
  //         path: 'food > fruits > bananas',
  //         isRefined: false,
  //         data: null
  //         count: 2,
  //       }, {
  //           name: 'kiwis',
[...]

PR #147, opened 2015-07-21
feat: hierarchical facets

vvo:
I have implemented the hierarchical facets as proposed in #137.

The available API is: (different from proposal)

Given records like this:

```json
{
  "name": "Flying dog easy IPA",
  "categories": {
    "lvl0": "beers",
    "lvl1": "beers > IPA",
    "lvl2": "beers > IPA > Flying dog"
  }
}
```

When using the helper:
```js
var helper = algoliasearchHelper(client, indexName, {
  hierarchicalFacets: [{
    name: 'categories', // mandatory
    attributes: ['categories.lvl0', 'categories.lvl1', 'categories.lvl2'] // mandatory, because we cannot guess what's the maximum level is (different from proposal)
    separator: '>' // optionnal, default to '>'
  }]
});

helper.toggleRefine('beers > IPA > Flying dog');

helper.on('result', function(content) {
  console.log(content.hierarchicalFacets);
  // [{
  //   'name': 'categories',
  //   'count': null,
  //   'isRefined': true,
  //   'path': null,
  //   'data': [{
  //     'name': 'beers',
  //     'path': 'beers',
  //     'count': 2,
  //     'isRefined': true,
  //     'data': [{
  //       'name': 'IPA',
  //       'path': 'beers > IPA',
  //       'count': 2,
  //       'isRefined': true,
  //       'data': [{
  //         'name': 'Flying dog',
  //         'path': 'beers > IPA > Flying dog',
  //         'count': 1,
  //         'isRefined': true,
  //         'data': null
  //       }, {
  //         'name': 'Anchor steam',
  //         'path': 'beers > IPA > Anchor steam',
  //         'count': 1,
  //         'isRefined': false,
  //         'data': null
  //       }]
  //     }]
  //   }]
  // }]
});

helper.search();
```

We always get the parent's current level values: if you refine on `beers > IPA > Flying dog`:
 - the hits will be only `Flying dog` beers
 - you will have a listing of all `Flying dog` beers and a listing of all `IPA` beers

I updated the original POC demo to use this new build (locally): it works.

Every reviewer gets one [Flying dog beer](http://flyingdogbrewery.com/beers/).

giphy-beer

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -12,7 +12,6 @@ var isUndefined = require('lodash/lang/isUndefined');
 var isString = require('lodash/lang/isString');
 var isFunction = require('lodash/lang/isFunction');
 var find = require('lodash/collection/find');
-var trimRight = require('lodash/string/trimRight');
 
 var extend = require('../functions/extend');
 var deepFreeze = require('../functions/deepFreeze');
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -755,6 +755,10 @@ SearchParameters.prototype = {
    * @return {SearchParameters}
    */
   toggleHierarchicalFacetRefinement: function toggleHierarchicalFacetRefinement(facet, value) {
+    if (!this.isHierarchicalFacet(facet)) {
+      throw new Error(facet + ' is not defined in the hierarchicalFacets attribute of the helper configuration');
+    }
+
     var separator = this.getHierarchicalFacetSeparator(this.getHierarchicalFacetByName(facet));
 
     var mod = {};
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -757,26 +757,24 @@ SearchParameters.prototype = {
   toggleHierarchicalFacetRefinement: function toggleHierarchicalFacetRefinement(facet, value) {
     var separator = this.getHierarchicalFacetSeparator(this.getHierarchicalFacetByName(facet));
 
-    return this.mutateMe(function merge(newInstance) {
-      var mod = {};
+    var mod = {};
 
-      // up one level:
-      // - `beer > IPA` => `beer` or
-      // - `beer` => ``
-      if (newInstance.hierarchicalFacetsRefinements[facet] === value) {
-        if (value.indexOf(separator) === -1) {
-          // root level
-          mod[facet] = '';
-        } else {
-          mod[facet] = trimRight(value.slice(0, value.lastIndexOf(separator)));
-        }
+    // up one level:
+    // - `beer > IPA` => `beer` or
+    // - `beer` => ``
+    if (this.hierarchicalFacetsRefinements[facet] === value) {
+      if (value.indexOf(separator) === -1) {
+        // root level
+        mod[facet] = '';
       } else {
-        mod[facet] = value;
+        mod[facet] = trimRight(value.slice(0, value.lastIndexOf(separator)));
       }
+    } else {
+      mod[facet] = value;
+    }
 
-      newInstance.hierarchicalFacetsRefinements = extend({}, newInstance.hierarchicalFacetsRefinements, mod);
-
-      return newInstance;
+    return this.setQueryParameters({
+      hierarchicalFacetsRefinements: extend({}, this.hierarchicalFacetsRefinements, mod)
     });
   },
   /**
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -11,6 +11,8 @@ var isEmpty = require('lodash/lang/isEmpty');
 var isUndefined = require('lodash/lang/isUndefined');
 var isString = require('lodash/lang/isString');
 var isFunction = require('lodash/lang/isFunction');
+var find = require('lodash/collection/find');
+var trimRight = require('lodash/string/trimRight');
 
 var extend = require('../functions/extend');
 var deepFreeze = require('../functions/deepFreeze');
@@ -80,6 +82,12 @@ function SearchParameters(newParameters) {
    * @member {string[]}
    */
   this.disjunctiveFacets = params.disjunctiveFacets || [];
+  /**
+   * All the declared hierarchical facets,
+   * a hierarchical facet is a disjunctive facet with some specific behavior
+   * @member {string[]|object[]}
+   */
+  this.hierarchicalFacets = params.hierarchicalFacets || [];
 
   // Refinements
   /**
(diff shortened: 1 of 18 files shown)


pixelastic:
I like it, it is quite easy to understand how this works. I have a few remarks and questions.

I think I understand why we need to repeat the previous parent categories in each `lvl` but this seems hackish. I understand that we cannot guess the priority order from the keys, but as a user I would expect a syntax with less clutter, like an `prioriies` key being an array of heading priority.

Something like:

```javascript
{
  "name": "Flying dog easy IPA",
  "categories": {
    "top_header": "beers",
    "subclass": "IPA",
    "final_page": "Flying dog"
  }
}

var helper = algoliasearchHelper(client, indexName, {
  hierarchicalFacets: [{
    name: 'categories',
    priorities: ['top_header', 'subclass', 'final_page'],
    separator: '>' // optionnal, default to '>'
  }]
});
```

Also, what is the `count` attribute in the results, and why is it `null`, shouldn't it be set to 2 ?

And finally I didn't get this comment:

> you will have a listing of all Flying dog beers and a listing of all IPA beers

If I filter on `beers > IPA > Flying dog`, won't I only have `Flying Dog` beers, so `IPA` only contains `Flying dog`?

vvo:
> I understand that we cannot guess the priority order from the keys, but as a user I would expect a syntax with less clutter, like an priorities key being an array of heading priority.

The thing is, what happens when you refine on `subclass: oranges` and your dataset have many `oranges` categories with the parent category being different?

You won't know if the user want "smoothies > oranges" or "fruits > oranges". And since this is disjunctive faceting, you would ask for "smoothies OR oranges" which is not what you want, I want the oranges smoothies.

Thus you need the full path at each category.

> If I filter on beers > IPA > Flying dog, won't I only have Flying Dog beers, so IPA only contains Flying dog?

You will get all the IPAS beers and all the Flying dog beers, because as a navigation system this is what you want:

```txt
| beers
  | IPA
    | Flying DOG
      - first flying dog
      - second flying dog
    | Belgian IPA
    | French IPA
```

You do get the listing for all IPA beers with the Flying dog opened, so that you can easily switch, otherwise the navigation is much harder for the end user.

Basically you always have the parent level values. We could change this default and hide under an option. Or the opposite, make it the default then provide an option to deactivate.

pixelastic:
Ok, I understand the `orange` example.

I don't understand how you managed the magic trick of getting the french and belgian IPA from a facetted search on Flying dog. Are you doing multiple queries?

vvo:
> Are you doing multiple queries?

Yes this is the same trick than for disjunctive facet, indeed hierarchical facet is just a disjunctive facet with a specific behavior.

There's one "hitQuery" and one "parentFacetValuesQuery" query for every disjunctive facet refinement.

See https://github.com/algolia/algoliasearch-helper-js/blob/7d31913150b4e1483028ca855aa7c53b7b3cbbd5/test/spec/hierarchical-facets/simple.js#L94-L95

pixelastic:
So what happen if you have for example 4 levels in your categories hierarchy ? Will you get facetted values for the parent category only, or facetted values for every categories excepte the top level one ? I guess this level of recursion could be defined as a config option

vvo:
> So what happen if you have for example 4 levels in your categories hierarchy ? Will you get facetted values for the parent category only, or facetted values for every categories excepte the top level one ? I guess this level of recursion could be defined as a config option

You get only the parent level values, let's wait for the usecase to popout before we implement this option?

vvo:
I will soon merge this, please comment if you are against it

bobylito:
In your example, you have a path like `beers > IPA > Flying dog` and the separator is `>`. What happens with the values? Do they contain spaces?

src/SearchParameters/index.js
```diff
@@ -726,6 +748,38 @@ SearchParameters.prototype = {
...
+   * Switch the refinement applied over a facet/value
+   * @method
+   * @param {string} facet name of the attribute used for facetting
+   * @param {value} value value used for filtering
+   * @return {SearchParameters}
+   */
+  toggleHierarchicalFacetRefinement: function toggleHierarchicalFacetRefinement(facet, value) {
+    var separator = this.getHierarchicalFacetSeparator(this.getHierarchicalFacetByName(facet));
+
+    return this.mutateMe(function merge(newInstance) {
```
bobylito: Why didn't you use the setQueryParameters method here instead of mutateMe?

src/SearchResults/generate-hierarchical-tree.js
```diff
@@ -0,0 +1,68 @@
...
+
+      while (level < currentHierarchicalLevel) {
+        parent = parent && find(parent.data, {isRefined: true});
+        level++;
+      }
+    }
+
+    // we found a refined parent, let's add current level data under it
+    if (parent) {
+      parent.data = sortByOrder(
```
bobylito: Pfoushh this `sortByOrder` just made my day ;)

src/SearchResults/index.js
```diff
@@ -218,56 +238,76 @@ function SearchResults(state, algoliaResponse) {
...
+          data: extend({}, dataFromMainRequest, facetResults),
+          exhaustive: result.exhaustiveFacetsCount
+        };
+        assignFacetStats(this.disjunctiveFacets[position], result.facets_stats, dfacet);
+
+        if (state.disjunctiveFacetsRefinements[dfacet]) {
+          forEach(state.disjunctiveFacetsRefinements[dfacet], function(refinementValue) {
+            // add the disjunctive refinements if it is no more retrieved
+            if (!this.disjunctiveFacets[position].data[refinementValue] &&
+              indexOf(state.disjunctiveFacetsRefinements[dfacet], refinementValue) > -1) {
```
bobylito: Isn't there an indentation problem here?

bobylito:
Looks good so far. I raised some stuff some questions though ;)

vvo:
> In your example, you have a path like beers > IPA > Flying dog and the separator is >. What happens with the values? Do they contain spaces?

No it will not contain spaces, values are trimmed.

We had a little talk with @bobylito, I did some magic and I did not explain why and how.

So to be more precise, if for a product you have `categories: {lvl0: 'beer', lvl1: 'beer > IPA'}` or `categories: {lvl0: 'beer', lvl1: 'beer>IPA'}` (notice, no space around `>`): Then the hierarchical result will be the same.

I do `trim()` after splitting on the separator `>`. This means that if you have spaces or not it does not matter.

The question is, should we handle this for ou users or not?

Pros:
- The user doesn't need to reformat his data or search the documentation for the separator whether he has spaces or no spaces around the separator

Cons:
- You cannot say that your categories separator is a ` ` (space character). Nor can it be a <TAB>)

Even if I do not like magic in programming, here it does brings something useful since I do not want to have support requests like: "I have weird spaces in my <a href=""></a> of the hierarchical menu, why?".

Also I should add that I originally did the examples and API proposal with a space around the separator because the POC I did with some client data was using spaces.

maxiloc:
@vvo Yes I would let the trim. But do you format it with a space for display ?

vvo:
after discussions with @maxiloc and @bobylito this need some more work, Ill update asap

vvo:
This PR is good to go.

tumblr_mrhfdlmusa1relg8bo1_250

bobylito:
MERGEEE

PR #151, opened 2015-07-23
fix: isDisjunctiveRefined() and isRefined() return true/false

### LINKED ISSUE #123: helper.isDisjunctiveRefined() returns true or undefined
Using `helper.isDisjunctiveRefined(name, value)` when the value is not refined, I get `undefined`.

I would expect to get either `true` or `false`.

Is this expected behavior?

If not, I will PR :-)

vvo:
instead of true/undefined
fixes #123

### DIFF
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -129,13 +129,16 @@ var lib = {
       refinementList[attribute].length > 0;
 
     if (isUndefined(refinementValue)) {
-      return containsRefinements;
+      return containsRefinements ? true : false;
+    }
+
+    if (!containsRefinements) {
+      return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/SearchParameters/RefinementList.js
```diff
@@ -129,13 +129,16 @@ var lib = {
       refinementList[attribute].length > 0;
 
     if (isUndefined(refinementValue)) {
-      return containsRefinements;
+      return containsRefinements ? true : false;
```
bobylito: Is this way of converting to boolean idiomatic?

I would go for the `!!var`.

vvo: Added the !!

bobylito:
:+1: ;)

PR #152, opened 2015-07-23
chore: debug scripts only in CI

vvo:
Using:

```bash
set -x
```

Only for CI (travis), because it outputs too much info for devs


bobylito:
Looks cool!

PR #153, opened 2015-07-23
Fix #126 : implement symetrical getState to complement setState

### LINKED ISSUE #126: Add a getState method
Since there is a setState method, it would be nice to also have a getState method that returns the current helper state.

bobylito:


vvo:
:+1:

PR #154, opened 2015-07-23
Fix #116 : a facet declared incorrectly should throw exceptions 

### LINKED ISSUE #116: Display warning msg when using disjunctive methods on regular facets
In the case I initialize regular facets, like that:
```
{
      ...
      params: {
        hitsPerPage: 4,
        facets: [
          'category',
          'brand'
        ]
      }
    }
```
And then use the wrong method `helper.addDisjunctiveRefine("category", "Electronics")`, instead of `helper.addRefine("category", "Electronics")` - I'd be great then to display a warning message in the JS console. Same in the opposite case.

bobylito:
For example if a facet is declared as disjunctive and we use `addRefine` with it then it should throw an exception.

### DIFF
--- a/test/spec/helper.distinct.js
+++ b/test/spec/helper.distinct.js
@@ -6,7 +6,10 @@ var forEach = require('lodash/collection/forEach');
 var algoliasearchHelper = require('../../index.js');
 
 test('Distinct not set', function(t) {
-  var helper = algoliasearchHelper(null, null, {});
+  var helper = algoliasearchHelper(null, null, {
+    facets: ['facetConj'],
+    disjunctiveFacets: ['facet']
+  });
   var state0 = helper.state;
 
   var disjunctiveFacetSearchParam = helper._getDisjunctiveFacetSearchParams();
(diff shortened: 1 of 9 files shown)


test/spec/helper.distinct.js
```diff
@@ -6,7 +6,10 @@ var forEach = require('lodash/collection/forEach');
 var algoliasearchHelper = require('../../index.js');
 
 test('Distinct not set', function(t) {
```
vvo: I do not understand why we add some more code the the helper distinct test while the PR is about throwing when unknown facet?

vvo:
Where do we test that it `throws`? t.throws

bobylito:
:( You're right :(

On Fri, Jul 24, 2015 at 12:26 PM, Vincent Voyer <notifications@github.com>
wrote:

> Where do we test that it throws? t.throws
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/algolia/algoliasearch-helper-js/pull/154#issuecomment-124466977>
> .
>



-- 
Alexandre Stanislawski
Software Engineer
Algolia

bobylito:
And now some test :)

vvo:
DONE!

PR #156, opened 2015-07-24
test: prefix integration index names with _travis-. Helps cleaning the indexes

vvo:


dethi:
@vvo why do you add `random` at the end?

vvo:
It seems travis build numbers may or may not be unique:
- https://github.com/travis-ci/travis-ci/issues/3028
- https://github.com/travis-ci/travis-ci/issues/2582

So I just add this little randomness in it

PR #159, opened 2015-07-24
Fix #150 : do not force distinct value to false if empty query or no filters

### LINKED ISSUE #150: Distinct forced to False
Hi I'm having a strange bug while making a simple request (first page of an listing) using Search Helper, even though I force `distinct` to be `true` it's overwritten to `false`.

Code:
```javascript
helper = algoliasearchHelper(client, INDEX, {
  facets: [ "recruiting" ],
  disjunctiveFacets: [ "branches", "city", "country", "employeesRange" ]
})
helper.setQueryParameter("distinct", true).search()
```

It seams to be connected with this line: https://github.com/algolia/algoliasearch-helper-js/blob/master/src/algoliasearch.helper.js#L559 (from the commit c8560a25a82313777e1c08d7a009eb0be913554f)

bobylito:


vvo:
LGTM

PR #175, opened 2015-07-28
chore: get rid of custom extend implementation, use lodash tools

vvo:


src/algoliasearch.helper.js
```diff
@@ -605,7 +605,7 @@ AlgoliaSearchHelper.prototype._getHitsSearchParams = function() {
     additionalParams.numericFilters = numericFilters;
   }
 
-  return extend(this.state.getQueryParams(), additionalParams);
+  return merge(this.state.getQueryParams(), additionalParams);
```
bobylito: Won't it change the behavior?

vvo: Should not, do you think about a failing case?

bobylito: Looking at the documentation, it doesn't seem like it could fail but it seems like it is a different and thus could bring unexpected results.

vvo: If we are able to write or think about a failing case then ok, right now I cannot think of one

bobylito: Ok let's go! :)

bobylito:
:+1: GOGOGO :)

vvo:
will do :)

PR #177, opened 2015-07-29
chore: ease the release and build process

vvo:
The build process contained:
- update version everywhere (package, bower, file)
- build
- build minify

Now:
- update version in CHANGELOG
- npm run release

The release process will also show you the build gzipped filesize.
So that if it gets to 500kb, we can investigate before releasing it.

The release process also check that we do have an upgrade in the version.
So that you cannot downgrade or mess with the version.

```sh
Build
..Minify
..Gzipped file size
algoliasearch.helper.min.js gzipped will weight 20.79 kB
```


bobylito:
LGTM :)

redox:
:+1:

PR #180, opened 2015-07-30
chore: fix integer annotations

vvo:


src/SearchParameters/index.js
```diff
@@ -272,7 +272,7 @@ function SearchParameters(newParameters) {
   /**
    * Enable the ranking informations in the response
    * @see https://www.algolia.com/doc#getRankingInfo
-   * @member {integer}
+   * @member {number}
```
bobylito: Could be worth checking if a boolean works here ;)

vvo: REST API says integer so we should keep it like that I guess https://www.algolia.com/doc/rest_api#getRankingInfo

bobylito: So we use number as it is the translation of the integer in JS?

vvo: Well since `{integer}` is not official, yes. But maybe we could just do the obvious: specify `{integer}` even if it's not a JS type. It still what the user needs to know.

We should feed the user with valuable information instead of trying to fit into jsdoc?

Or maybe a comment

bobylito: The user doesn't need to know about any other values than 1 (wich activates the configuration...). Imho {number} should be used to be consistent in our translation of the rest api into JS and we should add according comments like "set the value to 1 to activate the feature".

Actually jsDoc doesn't care at all about any type. It doesn't understand them, nor make any assumptions. Other tools might but not jsDoc itself.

vvo: I added the small comment

src/SearchParameters/index.js
```diff
@@ -4,7 +4,7 @@ var keys = require('lodash/object/keys');
 var intersection = require('lodash/array/intersection');
 var forEach = require('lodash/collection/forEach');
 var reduce = require('lodash/collection/reduce');
-var filter = require('lodash/collection/filter');
+var filter = require('lodash/collection/filter');e
```
bobylito: beware of the e ;)

vvo: how come the tests are passing :dancer:

PR #188, opened 2015-08-13
Fix #178 : getFacetValues

### LINKED ISSUE #178: API Proposal: Sorting facet values
following #38,

After discussing the API will be hosted by the SearchResults object. We will introduce 2 new functions to cover the needs for sorting and deprecate the previous `getFacetByName`.

```javascript
// RefinementValue is {name : string, count: number, isRefined : boolean }
results.getFacetValues('categories', {
  sortBy: ['isRefined:desc', 'count:desc'] // predicate list : string[] || sorting function :  (a: RefinementValue, b: RefinementValue
});
// Returns RefinementValue[]
// [ {name: 'skirt', count:'3', isRefined:true}, 
//   {name: 'jeans', count:'15', isRefined:false},
//   {name: 'pull', count:'10', isRefined:false} ]

results.getFacetStats('customerReviewCount');
// Returns
// {
//   "max": 7461,
//   "avg": 157.939,
//   "min": 1
// }
```

Checklist:

- [x] deprecate `getFacetByName`
- [x] implement `getFacetStats`
- [x] implement `getFacetValue`
- [x] update readme
- [x] update sample
- [x] document RefinementValue type

bobylito:
Adds a method so that the user can access the facet values by name with sorting built-in.

Also adds a method to get the stats.

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -528,7 +528,10 @@ SearchResults.prototype.getFacetValues = function(attribute, opts) {
 };
 
 /**
- *
+ * Returns the facet stats if attribute is defined and the facet contains some.
+ * Otherwise returns undefined.
+ * @param {string} attribute name of the facetted attribute
+ * @return {object} The stats of the facet
  */
 SearchResults.prototype.getFacetStats = function(attribute) {
   var facet = find(this.facets, {name: attribute});
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -32,6 +32,14 @@ var generateHierarchicalTree = require('./generate-hierarchical-tree');
  * @property {SearchResults.HierarchicalFacet[]} data sub values for the current level
  */
 
+/**
+ * @typedef SearchResults.FacetValue
+ * @type {object}
+ * @property {value} string the facet value itself
+ * @property {count} number times this facet appears in the results
+ * @property {isRefined} boolean is the facet currently selected
+ */
+
 function getIndices(obj) {
   var indices = {};
 
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -421,38 +421,38 @@ SearchResults.prototype.getFacetByName = function(name) {
 };
 
 /**
- * Get the facet values of a specified attribute from a SearchParameters object.
- * @param {SearchParameters} parameters the search parameters to search in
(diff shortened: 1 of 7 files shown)


bobylito:
All checks passed! Wanna review @redox :)

src/SearchResults/index.js
```diff
@@ -400,4 +420,122 @@ SearchResults.prototype.getFacetByName = function(name) {
...
 };
 
+/**
+ * Get the facet values of a specified attribute from a SearchResults object.
+ * @param {SearchResults} results the search results to search in
+ * @param {string} attribute name of the facetted attribute to search for
+ * @return {array|object} facet values enhanced with the attribute facetType to
+ * distinguish the type of facet.
+ */
+function getFacetValues(results, attribute) {
```
redox: I wondering which other name we could use here to avoid having the same name twice: `getFacetValues(SearchResults)` VS `SearchResults.prototype.getFacetValues`.

bobylito: I agree it's not nice but it is not exposed outside the module :)

test/spec/search.js
```diff
@@ -28,7 +28,45 @@ test('Search should call the algolia client according to the number of refinemen
...
+      {name: 'New York', count: 1, isRefined: true},
+      {name: 'San Francisco', count: 1, isRefined: false}
+    ];
+
+    t.deepEqual(
+      cityValues,
+      expectedCityValues,
+      'Facet values for "city" should be correctly ordered using the default sort');
+
+    var cityValuesCustom = data.getFacetValues('city', {sortBy: ['name:desc']});
```
redox: Could we also add a test sorting on `['count:asc', 'name:asc']` to ensure the magical lodash function is working?

bobylito: Sure

src/SearchResults/index.js
redox: Yeah got it, but still the first time I had trouble understanding who is doing what. Maybe `extract` or `find`?

bobylito: ok will change that to something more different / meaningful. What about extractNormalizedFacetValues? (since it normalize a the facet values for conjunctive and disjunctive, as well as extract them)

bobylito:
All clear ;)

redox:
LGTM :+1:

bobylito:
Great :D Mergin'

PR #189, opened 2015-08-13
FIX #136 : fire an event before the request is sent to Algolia

### LINKED ISSUE #136: Create an event triggered when a search request is sent

bobylito:


redox:
If the event is fired just **before** sending the query , the `requestsent` naming is maybe misleading. What about 'search' instead ?

bobylito:
Ok will fix



—
Alexandre STANISLAWSKI 
Software engineer 

Phone: +33 6 32 67 05 12 
Website: https://www.algolia.com/

Issue #148, opened 2015-07-21
Can't add array of NumericRefinement in addNumericRefinement


chrisyeung1121:
I have an array of codes like the following

``` javascript
$('.date-field').on('change', function() {
  helper.clearRefinements('occupied_at_i');
  var arrayOfDates = addDateFilters();
  _.forEach(arrayOfDates, function(n) {
  helper.addNumericRefinement('occupied_at_i', '!=', n);
  });
  helper.search();
});
```

I remembered once I tried I could get an array of values get meets the same operator but somehow I can't repeat it again. 

I tried: 

    helper.addNumericRefinement('occupied_at_i','!=',1233)
    helper.addNumericRefinement('occupied_at_i','!=',1234)

    helper.getRefinements('occupied_at_i')

And I got the following:

``` bash
Object
  operator: "!="
  type: "numeric"
  value: 1234
  __proto__: Object
```

Instead I want to get the following: 

``` bash
Object
  operator: "!="
  type: "numeric"
  value: [1233, 1234]
  __proto__: Object
```

What should I do?

Jerskouille:
I believe you're right, we seem currently [override the value][1].

[1]: https://github.com/algolia/algoliasearch-helper-js/blob/master/dist/algoliasearch.helper.js#L5685

redox:
Yes indeed, that's reported in https://github.com/algolia/algoliasearch-helper-js/issues/35. @bobylito could we work on that in the next few days?

bobylito:
It's on my todo!

chrisyeung1121:
Cheers!

bobylito:
Would it be ok if the API accepts arrays only for = and != operators? Or should we go for the same api that would construct a list for = and !=, while still keeping the same behaviour (override) for the other operators?

redox:
> Would it be ok if the API accepts arrays only for = and != operators? Or should we go for the same api that would construct a list for = and !=, while still keeping the same behaviour (override) for the other operators?

Even if I don't see right now a use-case for that, I assume it can happen (especially when you generate the filters programmatically). Anyway, I don't think it would change anything in the underlying implementation, isn't it?

Issue #35, opened 2015-04-02
Handle the whole scope of numeric filtering


bobylito:
The algolia API has the ability to set multiple values for a single parameter with or and and. https://www.algolia.com/doc#numericFilters

The helper api should be simple and not json based :P :palm_tree:

bobylito:
Example of friendly api : 

```javascript
helper.addNumericRefinement( "attr1", "=", [1,4,5,42] );
```

redox:
Looks good but it will not handle all the cases. 

Since the engine is able to handle ANDs or ORs: 
```
<or> ::= condition (OR condition)*
<filter> ::= <or> (AND <or>)*
```

Maybe we should define it like:
```
// attr1=42 AND (attr2=12 OR attr2=51)
helper.addNumericRefinement({ attribute: "attr1", operator: "=", value: 42 });
helper.addNumericRefinement([
  { attribute: "attr2", operator: "=", value: 12 },
  { attribute: "attr2", operator: "=", value: 51 }
]);
```

bobylito:
Thanks for the feedback. Let's iterate on that so we can make the most simple yet powerful api for handling those kind parameters.

reinaris:
:+1:

vvo:
```js
helper.addNumericRefinement({ attribute: "attr1", operator: "=", value: 42 });
helper.addNumericRefinement([
  { attribute: "attr2", operator: "=", value: 12 },
  { attribute: "attr2", operator: "=", value: 51 }
]);
```

The fact that the second call would mean `(attr2=12 OR attr2=51)` is really not obvious. In fact in most JS APIS this would just mean that the function accepts a single item or an array of items but not that they have different semantics.

While I can criticize the proposed API I do not have another solution for it right now.

pixelastic:
I can see the limitation of the current approach here too.

Maybe we could expose a high-level, SQL-like, string API to let user easily craft the query they'd like (eg. `attr1=42 AND (attr2=12 OR attr2=51)`) ? We'll transform that into our JSON-based representation of a query, while also letting more advanced users add to it directly.

On how to handle more complex cases while retaining a (more or less) readable structure, we can have a look at how things were done in other frameworks. Here's a [CakePHP example](http://book.cakephp.org/2.0/en/models/retrieving-your-data.html#complex-find-conditions).

This need more thoughts.

vvo:
[...]

PR #192, opened 2015-08-19
FIX #183 #148 #35 : Numeric filters fixed with support of advanced syntax

### LINKED ISSUE #183: API proposal : advanced numeric filters
**TL;DR : Algolia can do much more than what is actually possible with the helper! We need to refresh the API.**

## Updates of this issue
  - as suggested by @redox, the current behaviour of the add methods should be treated as a bug
  - if the user passes an array it will make OR predicates, otherwise the use of multiple add make them AND predicates

### On the Algolia side

The Algolia engine supports pretty advanced queries with two different type of predicates and let the user combine them with ANDs and ORs.

We have to two types of predicates that is supported : 
 - operator based predicates : a given attribute can be filtered with the use of an operator and a value. The supported operators are`=`, `!=`, `>`, `>=`, `<` and `<=`. For example, given that we have two numeric attributes a and b, we can do : 
 
```SQL
a!=3 AND (b=2 OR b>27)
```

  - range based predicates. No operators, just the range and the attribute. The single value is a special range with only value. They can be combined the same way as above. For example : 

```SQL
a:3 AND b:2 to 27 #this is not the same example
```

The last type of predicates can easily be emulated with operators.

### The helper side

The Helper provides an API to set operator based predicates with some limitations :
  - You can only combine predicates with ANDs
  - For a given pair of attribute-operator you can only set one value

Examples :
```SQL
# possible 
a=31 AND a!=22 AND b >2

# not possible /!\
a=3 AND a=2
a=3 OR b=42
```

## Enhancing the API

This proposition comes with certain aspects in mind :
  - be backward compatible as much as possible
  - fix inconsistencies
  - extend as much as we can without making the API too weird (lots of interpretation and personal gut feeling around this one :P)
  - old stuff should be as easy to do as before

### Extension of the current API

We have a set of methods (SearchParameters) to do operations on the numeric filters : add, remove, isRefined, get. They are either exposed or combined on the Helper : 

```javascript
// helper is already init
helper.addNumericRefinement('a', '=', 3); // Add the refinement
helper.addNumericRefinement('a', '=', 5); // Update the refinement=5
helper.getRefinement('a');  // returns {"=" : "3"}
helper.removeNumericRefinement('a', '='); // Remove the refinement for (a, =)
```

The current behavior is sort of a bug, as we use it to update. The semantic must change to ensure consistency with the other API and with the name itself. That's why the add will just add and not update anymore. 

```javascript
helper.addNumericRefinement('a', '=', 1); // Add the refinement a=1
helper.addNumericRefinement('a', '=', 2); // Add the refinement a=2
// at that point the resulting numericFilters=a:1,a:2  <=>  (a:1 AND a:2)
helper.getRefinement('a');  // returns {"=": [1, 2]}

helper.addNumericRefinement('a', '=', [3, 4]); // Add the refinement a=3 OR a=4
// at that point the resulting numericFilters=a:1,a:2,(a:3,a:4)  <=>  (a:1 AND a:2 AND (a:3 OR a:4))

helper.getRefinement('a');  // returns {"=": [1, 2, [3, 4]]}
```

### Introducing a raw API

The same way we can set the tags as algolia understand them in a string form : letting the user set the algolia parameter `numericFilters`. Previously this parameter was computed. The main characteristics are :
  - can't use both the other API and this one, to avoid collisions
  - no help is provided as for the handling or serializing those filters 
  - supports everything that Algolia could understand now or in the future
  
Example : 
```javascript 
helper.setQueryParameter("numericFilters", "a:3 to 5,(b : 2, b:5)");
helper.setQueryParameter("numericFilters", "a>=3, b<5,(b = 2, b = 5)");
```

## TODO

 - [x] Modify the current API
 - [x] Implement the advanced API
 - [x] Document the changes of the API in the README
 - [ ] Changelog

bobylito:


bobylito:
Ping @redox  :D

README.md
```diff
@@ -296,10 +296,17 @@ var helper = algoliasearchHelper(client, indexName, {
 });
 ```
 
-##### Add a numeric refinement
+##### Add numeric refinements
 
 ```js
 helper.addNumericRefinement('numericFacet', '=', '3').search();
```
redox: We need to replace `numericFacet` by `numericAttribute`, it's kind of misleading otherwise.

dist/algoliasearch.helper.js
```diff
@@ -7040,19 +7085,35 @@ SearchParameters.prototype = {
...
    * @method
    * @param {string} attribute attribute to set the filter on
    * @param {string} operator operator of the filter (possible values: =, >, >=, <, <=, !=)
    * @return {SearchParameters}
    */
-  removeNumericRefinement: function(attribute, operator) {
+  removeNumericRefinement: function(attribute, operator, paramValue) {
     if (!this.isNumericRefined(attribute, operator)) return this;
 
+    if (paramValue!==undefined) {
```
redox: Linting

redox:
LGTM :+1:

PR #198, opened 2015-09-02
fix: throw when unknown parameter

vvo:
Not sure we should merge this or should we wait for V3.. Because this
could break a good number of implementations..

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -325,6 +325,12 @@ function SearchParameters(newParameters) {
    * @member {string}
    */
   this.insideBoundingBox = params.insideBoundingBox;
+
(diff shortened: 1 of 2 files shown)


src/SearchParameters/index.js
```diff
@@ -325,6 +325,12 @@ function SearchParameters(newParameters) {
    * @member {string}
    */
   this.insideBoundingBox = params.insideBoundingBox;
+
+  forEach(params, function checkForUnknownParameter(paramValue, paramName) {
+    if (!this.hasOwnProperty(paramName)) {
+      throw new Error('algoliasearch-helper: unknown parameter: `' + paramName + '`');
```
redox: I would not prefix the error message with `algoliasearch-helper` (and we don't do it for other errors).

PR #207, opened 2015-09-08
chore: change the deprecated methods to avoid `members` in jsdoc

vvo:
jsdoc is showing deprecated methods as members (and thus at first in
the page).

This change will revert this and hide them from members (still shown
as methods, with deprecated indication).

2015-09-08-155359_721x656_scrot

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -138,7 +138,9 @@ AlgoliaSearchHelper.prototype.addDisjunctiveFacetRefinement = function(facet, va
 /**
  * @deprecated since version 2.4.0, see {@link AlgoliaSearchHelper#addDisjunctiveFacetRefinement}
  */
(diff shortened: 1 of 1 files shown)


vvo:
cc @pixelastic I think this was an issue for you to see members

bobylito:
LGTM

pixelastic:
image

This is what scared me :). This huge list of methods with obscure names like `addExclude`, `addFacetExclusion`, `removeFacetRefinement`, `clearRefinements`, `hasRefinements`, `isRefined`. I plan to pass on each one to see exactly how it works, with jsfiddle/codepen examples.

But removing members is good too ;)

vvo:
> This is what scared me :).

A lot of them are deprecated and we must find a way to put them at the end of the list in a "deprecated" with red color/low gamma.

Dunno how to do that, we have to dig in the jsdoc thing cc @bobylito :dancer:

Issue #203, opened 2015-09-08
Provide a set of utility methods to help handling URLs


bobylito:
As we continue the development of the instantsearch.js, we need to provide a set of utility methods on the helper to make saving and restoring a search state easily (see [related issue in the instant search project](https://github.com/algolia/instantsearch.js/issues/17)). We will need to provide 3 methods on the helper : 

 - one that creates a plain javascript object with only the specified subset of parameters from the current search state. This one could be implemented as specific behavior of the current getState method.
 - two that will serialize and deserialize a given subset of parameters and transform them into a query string form

The format to specify the subset is : 
```js
[
  // Any parameters of the SearchParameters can be chosen. Most likely to be chosen : 
  "query",
  "page",
  "hitsPerPage",
  // Also the index, as it can be changed for sorting purpose but is not in the SearchParameters
  "index",
  // Facets can be set through the usage of different widgets. But the user don't know much about that, only that it is based on an attribute
  "attribute:myAttribute",
  // For simplicity sake, we can choose to track all the facets.
  "attribute:*"
]
```

## API proposal

```js
var filters = ["numeric:price", "facet:category", "query"]
helper.getState(filters);
helper.getStateAsQueryString(filters, moreAttributes);
helper.setStateFromQueryString(queryString, filters);
```

Update : remove the queryString() factory method and use only plain methods

pixelastic:
I feel it counter-intuitive that `helper.queryString()` returns an object and not a real query string. I would see more something like a `helper.exportStateToQueryString(filters)` and `helper.setStateFromQueryString(queryString)` instead

vvo:
Great that we (YOU :D) are going to do this!

> helper.queryString().serialize(filters, moreAttributes);
What is `moreAttributes`?

> helper.queryString().deserialize(serializedString, filters);

Why do you have a first step being `.queryString()`? Will we be able to somehow use its result. If so what would it look like?

Also if `.queryString()` is a sort of first step then I thought it would accepts `.queryString(filters..)`?

As for the API proposal, what about:

```js
helper.getStateAsQueryString(filters);
helper.setStateFromQueryString(string, filters);
helper.getState(filters);
```

?

Maybe you need more than that for other use cases, you tell me.

> The format to specify the subset is :
> 
> [
>   // Same goes with numerics (because they are not actual facets)
>   "numeric:*",
>   "numeric:price"
> ]

[...]

PR #209, opened 2015-09-10
Url helpers

bobylito:
New methods to make url synchronisation implementation easier. It contains 3 methods : 
  - AlgoliaSearchHelper.setState(filters) to get a filtered state (this is an enhancement of the current implementation)
  - AlgoliaSearchHelper.getStateAsQueryString to get a part of the state as a queryString
  - AlgoliaSearchHelper.setStateAsQueryString to set the state of the helper with a query string

With those methods, the helper user has the ability to create and read a custom query string that he can then use to update the URL. It is possible to pass an object for custom attributes not handled by the helper (existing values in the webpage). To avoid name collisions, it is also possible to prefix the attributes.

```js
  var initialState = {
    facets: ['facetA', 'facetWeDontCareAbout'],
    disjunctiveFacets: ['facetB']
  };
  var index = 'index';
  var helper = algoliasearchHelper(null, index, initialState);

  helper.setQuery('a query');
  helper.toggleRefine('facetA', 'a');

  var filters = ['query', 'attribute:facetA']; //attribute can also be set to * to catch all the attributes

  var partialState = helper.getState(filters);
  /*
{
  "facetsRefinements": {
    "facetA": [
      "a"
    ]
  },
  "query": "a query"
}
  */
  var queryString = helper.getStateAsQueryString(filters);
/* facetsRefinements%5BfacetA%5D%5B0%5D=a&query=a%20query */
  var qs2 = helper.getStateAsQueryString(filters, {
    prefix: 'prfx_',
    moreAttributes: {foo: 'bar'}
  });
/* prfx_facetsRefinements%5BfacetA%5D%5B0%5D=a&prfx_query=a%20query&foo=bar */

  // When instanciating a new helper, we must provide informations that are not saved in the query string
  var newHelper = algoliasearchHelper(null, null, initialState); 
  newHelper.setStateAsQueryString(queryString);
/*
{
  "query": "a query",
  "facets": [
    "facetA",
    "facetWeDontCareAbout"
  ],
  "disjunctiveFacets": [
    "facetB"
  ],
  "hierarchicalFacets": [],
  "facetsRefinements": {
    "facetA": [
      "a"
    ]
  },
  "facetsExcludes": {},
  "disjunctiveFacetsRefinements": {},
  "numericRefinements": {},
  "tagRefinements": [],
  "hierarchicalFacetsRefinements": {},
  "page": 0
}
*/
```

Update : I also added two functions on the AlgoliaSearch object to read query strings and get the helper state/configuration and the parameters of other apps/libs.
Update 2 : made the helper parse the string numeric values into numbers.
Update 3 : index is now a part of the searchParameters


bobylito:
Ping @vvo @pixelastic @redox :)

redox:
I think we spoke about parameters shortening, like `<PREFIX>-atr=` instead of `<PREFIX>-attributesToRetrieve` -> is that still something we're gonna add?

src/algoliasearch.helper.js
```diff
@@ -431,10 +443,118 @@ AlgoliaSearchHelper.prototype.setState = function(newState) {
...
+  var partialState = this.getState(filtersOrDefault);
+  var partialStateWithPrefix = prefixForParameters ?
+    mapKeys(
+      partialState,
+      function(v, k) { return prefixForParameters + k; }) :
+    partialState;
+
+  if (moreAttributes) merge(partialStateWithPrefix, moreAttributes);
+
+  return qs.stringify(partialStateWithPrefix);
```
redox: What about using `qs.stringify(partialStateWithPrefix, { arrayFormat: 'repeat' })` to ease the readability (but we'll need to have shortest query parameter names)?

Also, I'm wondering if we should also handle the `strictNullHandling` option?

CHANGELOG
```diff
@@ -1,5 +1,12 @@
 CHANGELOG
 
+2.4.0 - 2015-09-10
+  * FEATURE : add AlgoliaSearchHelper.setState(filters) to get a filtered state
```
redox: `s/setState/getState/`

src/algoliasearch.helper.js
```diff
@@ -431,10 +443,118 @@ AlgoliaSearchHelper.prototype.setState = function(newState) {
...
+  return partialState;
+};
+
+/**
+ * Get part of the
  */
-AlgoliaSearchHelper.prototype.getState = function() {
-  return this.state;
+AlgoliaSearchHelper.prototype.getStateAsQueryString = function getStateAsQueryString(filters, options) {
+  var moreAttributes = options && options.moreAttributes;
```
redox: Do we really need to nest the "extra attributes", what about taking all the keys in `options` (`prefix included anyway)? Then I would rename `options` to `additionalParameters`.

pixelastic:
I would rename `setStateAsQueryString` to `setStateFromQueryString`.

Also, didn't we suggest we should wrap all our data in only one query string variable? Something like `page=faq&agolia_unique_var={...}`. With `page` one of the query string variables already defined on the website, `algolia_unique_var` something unique enough to do not conflict with the underlying website and `{...}` and encoded version of our variables (escaping `&`, `=`, etc).

pixelastic:
Oh, just saw your comment on #203. Following the discusion there.

bobylito:
> I would rename setStateAsQueryString to setStateFromQueryString.

Ok it'll be done!

> Also, didn't we suggest we should wrap all our data in only one query string variable? Something like page=faq&agolia_unique_var={...}. With page one of the query string variables already defined on the website, algolia_unique_var something unique enough to do not conflict with the underlying website and {...} and encoded version of our variables (escaping &, =, etc).

This way has many draw backs : 
 - there will be a double encoding of the characters, which will make the query string very long
 - it obfuscates the data even more
 - it's an edge case, anyway

pixelastic:
I answered in #203, but let's continue here :)

- I don't see the double encoding of characters, we'll only have to encode all `&`, `?` in our data so it does not confuse the rest of the querystring.
- Agreed that it obfuscated a bit more the data, but the url is not supposed to be manually edited.
- I als don't see the edge case, people will have all sorts of variables in their urls and we must not mess with them. We should have the smaller footprint possible in term of namespacing in the url (ie. use only one entrypoint).

vvo:
> /* facetsRefinements%5BfacetA%5D%5B0%5D=a&query=a%20query */

@bobylito It should not output encoded query string by default, the decision to encode it is on the user side, it all depends where he will use it.

If we use it inside the url (location.pushState..) then there's no need to encode it, so that it will look a lot better in the url.

This is what I talked you about, I dunno why hapijs/qs encodes by default because there's no need to do so.

So we need to decodeURIComponent what hapijs sends us before returning the string.

I hope my explanation is clear enough.

vvo:
In fact there's never a need to encodeURIComponent BUT if you store the result INSIDE a queryString parameter (something like aa-url=FULLYENCODEDURL). Otherwise there's no need to do so.

vvo:
> I think we spoke about parameters shortening, like <PREFIX>-atr= instead of <PREFIX>-attributesToRetrieve -> is that still something we're gonna add?

I would really love to see this from the start of this feature. Having aliases right away and good defaults would help make our url clean and short.

```
getStateAsQueryString({
  filters,
  aliases: {
    disjunctiveFacetsRefinements: 'dfr',
    ...
  },
  prefix
  additionnalAttributes
})
```

Thoughts?

vvo:
@bobylito What do you think of `additionalAttributes` instead of `moreAttributes`?

vvo:
@bobylito Since filters and params are optional in getStateAsQueryString, why not accept directly an otion object?

Instead of:

```js
getStateAsQueryString(filters, {moreAttributes, prefix});
```

I would do:

```js
getStateAsQueryString({filters, moreAttributes, prefix});
```

Since filters is optional (which is different from something like `client(appId, apiKey)` where those are mandatory)

vvo:
> I als don't see the edge case, people will have all sorts of variables in their urls and we must not mess with them. We should have the smaller footprint possible in term of namespacing in the url (ie. use only one entrypoint).

Yes for this, I fear we can get some good support tickets because of this cool feature (urlSync). What @redox proposed (autoprefix + alias) seems to be a good solution here.

vvo:
> I would rename setStateAsQueryString to setStateFromQueryString.

:+1:

src/algoliasearch.helper.js
bobylito: Why not, just think it's a bit odd to pass the configuration not just the data.

CHANGELOG
```diff
@@ -1,5 +1,12 @@
 CHANGELOG
 
+2.4.0 - 2015-09-10
```
vvo: I would set as `UNRELEASED` as we do in PRS because it's hard to know when it will be really released when in the PR

vvo:
> attribute:facetA

Why not `attributes` since it's a list?

src/algoliasearch.helper.js
vvo: :+1: 

https://github.com/hapijs/qs#stringifying

```
Qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'indices' })
// 'a[0]=b&a[1]=c'
Qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'brackets' })
// 'a[]=b&a[]=c'
Qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'repeat' })
// 'a=b&a=c'
```

Did not know about it.

vvo: As array query string is not standardized, I dunno what is the support for these options.

vvo: > Also, I'm wondering if we should also handle the strictNullHandling option?

Dunno also if we have situations where we use null as a token and expect some behavior from it.

test/spec/helper.state.js
```diff
@@ -4,6 +4,8 @@ var test = require('tape');
 var algoliasearchHelper = require('../../index');
```
vvo: WEeee tests! :heart_eyes:

vvo:
Travis failed because.. It's using node v4 now. Will have a look.

CHANGELOG
bobylito: Ok!

bobylito:
> Since filters is optional

Filters are not optional right now...

vvo:
> Filters are not optional right now...

I did not know that, I thought that by default we would just get all the relevant (needed) parameters.

I think it would make sense to make this function have a good default behavior without having to specify filters.

bobylito:
> by default we would just get all the relevant (needed) parameters.

What are the relevant parameters? I think it's pretty hard to guess, and very easy to specify.
**But** we can easily have a default, that just works like `['index', 'query', 'attribute:*']`. What do you think?

bobylito:
For the parameter shortening, since we have lots of parameters that could go in the url (all the state actually), what about we do that programmatically? Maybe using the the first letter and the following uppercased letters? For example : `query` will be `q` and `tagRefinements` will be `tr`

vvo:
> For the parameter shortening, since we have lots of parameters that could go in the url (all the state actually), what about we do that programmatically? Maybe using the the first letter and the following uppercased letters? For example : query will be q and tagRefinements will be tr

Sounds great, curious to see if it works in all cases but if so cool because you can precompute the full aliases in an object (to be reused when deserializing).

vvo:
> But we can easily have a default, that just works like ['index', 'query', 'attribute:*']. What do you think?

Would be cool yes!

src/algoliasearch.helper.js
```diff
@@ -431,10 +439,184 @@ AlgoliaSearchHelper.prototype.setState = function(newState) {
...
+  var prefixForParameters = options && options.prefix || '';
+
+  var partialStateWithPrefix = qs.parse(queryString);
+  var prefixRegexp = new RegExp('^' + prefixForParameters);
+  var partialState = mapKeys(
+    partialStateWithPrefix,
+    function(v, k) {
+      if (prefixForParameters && prefixRegexp.test(k)) {
+        var encodedKey = k.replace(prefixRegexp, '');
+        return shortener.decode(encodedKey);
```
redox: Missing `|| encodedKey` ?

src/algoliasearch.helper.js
```diff
@@ -431,10 +439,184 @@ AlgoliaSearchHelper.prototype.setState = function(newState) {
...
+  var filters = options && options.filters || ['query', 'attribute:*'];
+  var moreAttributes = options && options.moreAttributes;
+  var prefixForParameters = options && options.prefix || '';
+
+  var partialState = this.getState(filters);
+
+  var encodedState = mapKeys(
+    partialState,
+    function(v, k) {
+      var shortK = shortener.encode(k);
```
redox: missing `|| k` ?

src/algoliasearch.helper.js
```diff
@@ -431,10 +439,184 @@ AlgoliaSearchHelper.prototype.setState = function(newState) {
...
+        });
+        numericRefinements[attribute][operator] = parsedValues;
+      });
+    });
+    partialState.numericRefinements = numericRefinements;
+  }
+
+  if (partialState.page) {
+    partialState.page = parseInt(partialState.page, 10);
+  }
```
redox: Why do we need to handle the page & the numeric refinements differently?

src/algoliasearch.helper.js
bobylito: If the key match the prefix (which is supposed to be a namespace for our keys) and is not decodable, it means that the prefix was not good. What about we throw an error?

src/algoliasearch.helper.js
bobylito: Are you referring to parsefloat va parseint?

redox: No no, but I was wondering why is this function dealing with the numericRefinements & page and not with all other parameters.

src/algoliasearch.helper.js
bobylito: We have a test to check the exhaustivity of the shorten versions of the keys, so we should have a shorten version for each key in the searchparameters.

src/algoliasearch.helper.js
bobylito: Will make that exhaustive :)

bobylito:
Updated the PR, should be good for review.

package.json
```diff
@@ -78,7 +78,8 @@
   },
   "dependencies": {
     "lodash": "^3.10.1",
-    "lodash-compat": "^3.10.1"
+    "lodash-compat": "^3.10.1",
+    "qs": "5.1.0"
```
vvo: We can (and should) now use qs: ^5.1.0. I know this is the opposite of what we decided some months ago but this is in the end the best way to avoid duplicate dependencies for front end developers when they do build with browserify and webpack..

We now have the shinkwrap to handle our build the right way (reproducible).

vvo: npm config set save-exact false

src/SearchParameters/index.js
```diff
@@ -333,12 +342,68 @@ function SearchParameters(newParameters) {
 
   forOwn(params, function checkForUnknownParameter(paramValue, paramName) {
     if (!this.hasOwnProperty(paramName)) {
-      console.error('Unsupported SearchParameter: `' + paramName + '` (this will throw in the next version)');
+      /*eslint-disable*/
```
vvo: Should not be needed since console and line length are warnings in our eslint configuration, maybe you want to avoid warnings somehow?

CHANGELOG
```diff
@@ -1,5 +1,18 @@
...
+  * FEATURE : add AlgoliaSearchHelper.getState(filters) to get a filtered state
+  * FEATURE : add AlgoliaSearchHelper.getStateAsQueryString to get a part of
+  the state as a queryString
+  * FEATURE : add AlgoliaSearchHelper.setStateFromQueryString to set the state
+  of the helper with a query string
+  * FEATURE : add AlgoliaSearchHelper.getConfigurationFromQueryString. Reads
+  a query string and returns the associated partial helper configuration
+  * FEATURE : add AlgoliaSearchHelper.getForeignConfigurationInQueryString to
+  get the configuration contained in a qs that is not from the helper
+  * FIX : all the properties that are supposed to be numbers are now parsed in
```
vvo: Can you give more details on why we needed to do so?

src/SearchParameters/shortener.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+var invert = require('lodash/object/invert');
+
+var keys2Short = {
```
vvo: :heart_eyes:

test/spec/helper.clears.js
```diff
@@ -35,7 +35,7 @@ test('Check that the state objects match how we test them', function(t) {
   t.deepEqual(helper.state.facetsExcludes, {excluded1: ['0'], excluded2: ['0']});
   t.deepEqual(
     helper.state.numericRefinements,
-    {numeric1: {'>=': ['0'], '<': ['10']}, numeric2: {'>=': [0], '<': [10]}});
+    {numeric1: {'>=': [0], '<': [10]}, numeric2: {'>=': [0], '<': [10]}});
```
vvo: Hope this will not break anything :D

vvo:
LGTM, very very nice to have all this in the helper

CHANGELOG
bobylito: Consistency. We were able to have them in both format (string and numbers) but when deserializing would always use strings. Therefore we had inconsistencies when testing with numbers as input and we wanted to compare the results of the values.

This way we are guaranteed that the numeric will be numeric.

package.json
bobylito: Ok! (thanks for the tricks :) )

CHANGELOG
vvo: Ok cool nice!

README.md
```diff
@@ -656,6 +656,42 @@ var state1 = helper.searchOnce({hitsPerPage: 1})
 });
 ```
 
+### URL Helpers
+
+#### Get a query string of a part of the parameters
+
+```js
+// see the jsdoc for the parameters
```
vvo: Maybe link to the jsdoc because here I am frustrated right now! :D

src/SearchParameters/index.js
bobylito: Warnings are to be avoided the same way the errors are! We ought to remove the warning through the rules or in the code. I did it in the code.

PR #211, opened 2015-09-10
fix: getFacetStats should look in both facets, disjunctiveFacets

vvo:


src/SearchResults/index.js
```diff
@@ -534,8 +534,17 @@ SearchResults.prototype.getFacetValues = function(attribute, opts) {
...
+
+  if (this._state.isConjunctiveFacet(attribute)) {
+    facet = find(this.facets, {name: attribute});
+    return facet && facet.stats;
+  } else if (this._state.isDisjunctiveFacet(attribute)) {
+    facet = find(this.disjunctiveFacets, {name: attribute});
+    return facet && facet.stats;
+  }
+
+  throw new Error(attribute + ' is not present in `facets` or `disjunctiveFacets`');
```
vvo: Changing the default behavior for undeclared facets, I guess that's better but wonder if it can break some implementations.

bobylito: at least there is a message telling the user there is something wrong ;)

bobylito:
LGTM

bobylito:
Mergin' (forgot to merge this PR :/)

PR #212, opened 2015-09-11
FIX #208, getFacetValues with no values for the facet should not throw.

### LINKED ISSUE #208: getFacetValues throws when no hits returned
When a search returns no hits, calling `getFacetValues` for an existing facet will throw in the `extractNormalizedFacetValues ` function.

At [this line](https://github.com/algolia/algoliasearch-helper-js/blob/master/src/SearchResults/index.js#L434), `facet` is undefined because the `results.facets` array is empty, so trying to access `data` on `facet` raises an error:

```javascript
    if (results._state.isConjunctiveFacet(attribute)) {
        var facetValues = map(facet.data, function(v, k) { // Uncaught TypeError: Cannot read property 'data' of undefined
```

I guess the problem would arise on lines [445](https://github.com/algolia/algoliasearch-helper-js/blob/master/src/SearchResults/index.js#L445) and [456](https://github.com/algolia/algoliasearch-helper-js/blob/master/src/SearchResults/index.js#L456) as well!

While it's easy enough to check for `nbHits` before calling `getFacetValues`, I guess it'd be better to safeguard the function internally :)

bobylito:
Instead it should gracefully return an empty array (the behavior was
correctly defined for the hierarchical facets)

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -420,6 +420,8 @@ SearchResults.prototype.getFacetByName = function(name) {
     find(this.hierarchicalFacets, predicate);
 };
 
+var EMPTY_NORMALIZED_FACET_VALUES = (function() { var a = []; a.facetType = 'conjunctive'; return a; })();
+var EMPTY_NORMALIZED_DISJUNCTIVE_VALUES = (function() { var a = []; a.facetType = 'disjunctive'; return a; })();
 /**
  * Get the facet values of a specified attribute from a SearchResults object.
  * @param {SearchResults} results the search results to search in
@@ -431,6 +433,8 @@ function extractNormalizedFacetValues(results, attribute) {
   var predicate = {name: attribute};
   if (results._state.isConjunctiveFacet(attribute)) {
     var facet = find(results.facets, predicate);
+    if (!facet) return EMPTY_NORMALIZED_FACET_VALUES;
+
     var facetValues = map(facet.data, function(v, k) {
       return {
         name: k,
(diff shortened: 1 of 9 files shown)


bobylito:
@redox

redox:
LGTM :+1:

src/SearchResults/index.js
```diff
@@ -420,6 +420,8 @@ SearchResults.prototype.getFacetByName = function(name) {
     find(this.hierarchicalFacets, predicate);
 };
 
+var EMPTY_NORMALIZED_FACET_VALUES = (function() { var a = []; a.facetType = 'conjunctive'; return a; })();
```
vvo: Why do we need to attach a property containing `facetType` to the array here? why not just return an empty array?

vvo: Also what do you think of:

```js
var EMPTY_NORMALIZED_FACET_VALUES = [];
EMPTY_NORMALIZED_FACET_VALUES.facetType = 'conjunctive'
``

"Ugly" But easy to read.

vvo: > Why do we need to attach a property containing facetType to the array here? why not just return an empty array?

Ok I saw the other lines in the helper, I did not know we had this in the helper. Is it something we use in some implementations? (facetType)

bobylito: Actually, after reading the code, this property is only necessary for case it's a hierarchical facet. We may remove it :)

vvo: Damn hierarchicalFacets! Who did this!!? :dancer:

vvo:
Lint did not pass

bobylito:
Fixing :)

src/SearchResults/index.js
bobylito: But the documentation states that this values is attached when returned from extractNormalizedFacetValues, but then we should probably remove it from the returned array.

vvo: Did not even know about extractNormalizedFacetValues O_O

vvo:
FLGTM

bobylito:
Everything is alright then, merging!

PR #217, opened 2015-09-16
chore: stop using pinned deps

vvo:
Our modules can be used in two ways:
- direct npm dependent
- `<script>` on jsDelivr

We were previously using "pinned" dependencies (1.2.3), I have
switched them to ^1.2.3, mostly to avoid code duplication.

To tackle code duplication in modules using us as a direct
dependency, I switched all direct dependent to ^x.x.x version.

This means "We are compatible with any minor/patch version you may
already have for this dependency". This allow the host project(s) to
avoid code duplication.

As for the builded files, I added an npm-shrinkwrap.json file to lock
down our dependencies.

This has a very nice advantage that now we do control the result of
the build being pushed on jsdelivr. While previously it was up to the
developer node_modules state.

Since we have an npm run release task and since it's doing `npm
install`, we will always end up with a build where the dependencies
are locked.

The dev dependencies stays the same because there's no clear win to
auto upgrade them BUT break the build :)


vvo:
cc @bobylito

bobylito:
What if the dependency breaks our work?

vvo:
> What if the dependency breaks our work?

Then we are screwed.

This is a tradeoff, either every host project (client project) will have lodash duplicated because they are using "3.0.1" and we are using "3.0.2" either we do like this PR suggests and then it could happen that we break.

BUT this is only true for frontend projects built using npm/browserify/webpack and any module loader that are not using shrinkwrap.

When doing a frontend project with lots of dependencies and building it, if you are not using shrinkwrap in your project you will have issues, not just with Algolia. Trying to solve this use case is not something we need to do. People need to use shrinkwrap.

bobylito:
LGTM

PR #219, opened 2015-09-23
fix:IE8 has no array indexof

vvo:


vvo:
cc @bobylito

bobylito:
LGTM.

There is a formatting problem on src/SearchParameters/index.js 348

vvo:
fixed the lint error

PR #222, opened 2015-09-24
fix(peerDependencies): Removes algoliasearch from the peerDependencies

pixelastic:


bobylito:
Why?



—
Alexandre STANISLAWSKI 
Software engineer 

Phone: +33 6 32 67 05 12 
Website: https://www.algolia.com/

pixelastic:
Sorry, tried @vvo `gh` tool to submit the PR and the description got deleted in the process. Also, wrong branch. Will submit a new one.

vvo:
Indeed @bobylito we talked briefly with @pixelastic and since nowhere in the helper we are using require('algoliasearch'), it means it's not a peerDependency. You have to pass the client to the helper so you will need it of course but no need to set it as a peerDependency.

vvo:
The helper is dependent on algoliasearch instances, not on the module itself

bobylito:
Cool makes sense.

PR #223, opened 2015-09-24
fix(deps): Removes `algoliasearch` from the dependencies.

pixelastic:
We never `requires` it, we expect it passed as an argument, so we
should be able to safely remove it from the dependencies. It is not
needed to build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,8 +80,5 @@
     "lodash": "^3.10.1",
     "lodash-compat": "^3.10.1",
     "qs": "5.1.0"
-  },
-  "peerDependencies": {
-    "algoliasearch": ">= 3.1 < 4"
   }
 }


pixelastic:
As said in the previous PR (that I wrongly pushed to master, sorry for the noise), we do not really have `algoliasearch` as a dependency here, we just expect it to be passed as an argument.

bobylito:
Well actuall this is exactly what peer dependencies are for. https://nodejs.org/en/blog/npm/peer-dependencies/ From the reading of this article, we can say that the helper is sort of in the same situation. It does use the package but actually can't live without it. It's also good as it serve as a compatibility reference.

vvo:
Did not read the entire of the article but the header says "The problem: plugins" And indeed the helper is not a plugin of the client.

This issue came because npm warns that it will no more install peerDependencies in npm3 etc.. Thus making noise for something that seems not to be like a plugin.

You decide!

bobylito:
Plugins is an example. At the end of the day there is a dependency between the helper and the client. I'm glad NPM3 won't install this as a dependency but AlgoliaSearch must be present if you are making a project with helper. Without that npm won't be able to figure out there is a missing dependency.

pixelastic:
After reading the article, I actually better understand the `peerDependencies`, and that is something we should keep. Too bad npm warns us of that "issue"

PR #224, opened 2015-09-28
chore(shortener): Sort key names alphabetically

pixelastic:
Just a sorting of the shortener keys alphabetically. Makes searching for an attribute easier and will make diffs cleaner.


bobylito:
LGTM

vvo:
Is this linked with the instantsearch issue about keeping the same order than what was defined in filters:[]?

bobylito:
No :)

pixelastic:
Nope, I just had to look at this file to choose a key that wasn't used for the current instantsearch.js version that is now stored in the url. I just applied the boy-scout rule here (_leave the code better than you found it_)

PR #229, opened 2015-10-10
fix(facetStats): a facet can be both in facets & disjunctiveFacets

### LINKED ISSUE #228: When adding an attribute in both the facets and disjunctiveFacets, you cannot retrieve the stats
If you initialize the helper with:

```js
{
  facets: ['price'],
  disjunctiveFacets: ['price']
}
```

You cannot anymore get the `getFacetStats('price')` because both `isConjunctiveFacet` and `isDisjunctiveFacet` are `true` (expected) but the underlying `facet` object is only in one of them (in `disjunctiveFacets`).

redox:
Fix #228


redox:
@bobylito do you understand what's going on?

bobylito:
LGTM :)

PR #231, opened 2015-10-13
feat(urlTools) : expose all url tools and more in a dedicated module

bobylito:
- new namespace `url` for all functions related to query string serialization deserialization
 - move static function to the new module
 - added a new function to serialize any `SearchParameters`
 - getState filter feature now exposed as filter method on `SearchParameters`

### DIFF
--- a/README.md
+++ b/README.md
@@ -684,15 +684,21 @@ helper.getState(['query', 'attribute:*']);
 #### Get an object for the helper configuration in the query string
 
 ```js
-var state = algoliasearchHelper.AlgoliaSearchHelper.getConfigurationFromQueryString(qs);
+var state = algoliasearchHelper.url.getStateFromQueryString(qs);
 ```
 
 #### Get the configuration contained in the query string that is not for the helper
 
 ```js
-var otherConf = algoliasearchHelper.AlgoliaSearchHelper.getForeignConfigurationInQueryString(qs);
+var otherConf = algoliasearchHelper.url.getUnrecognizedParametersInQueryString(qs);
 ```
 
+#### Get the query string of any state
+
+```js
+var state = helper.state;
+var qs = algoliasearchHelper.url.getQueryStringFromState(state);
+```
 
 ### Query parameters
 
--- a/test/spec/compat.js
+++ b/test/spec/compat.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var test = require('tape');
+var algoliaSearchHelper = require('../../index');
+var url = algoliaSearchHelper.url;
+var AlgoliaSearchHelper = algoliaSearchHelper.AlgoliaSearchHelper;
+
+test('Should be compatible', function(t) {
+  t.equal(
+    AlgoliaSearchHelper.getConfigurationFromQueryString,
+    url.getStateFromQueryString
+  );
+  t.equal(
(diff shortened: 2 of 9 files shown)


redox:
Nice :)

I would go for a few renamings:
 - `getConfigurationFromQueryString -> getStateFromQueryString`
 - `getForeignConfigurationInQueryString -> getUnrecognizedParametersInQueryString`
 - not yet sure about `urlTools`, maybe I would go for just `url`

bobylito:
All done @redox!

README.md
```diff
@@ -684,15 +684,21 @@ helper.getState(['query', 'attribute:*']);
...
 ```js
-var state = algoliasearchHelper.AlgoliaSearchHelper.getConfigurationFromQueryString(qs);
+var state = algoliasearchHelper.url.getStateFromQueryString(qs);
 ```
 
 #### Get the configuration contained in the query string that is not for the helper
 
 ```js
-var otherConf = algoliasearchHelper.AlgoliaSearchHelper.getForeignConfigurationInQueryString(qs);
+var otherConf = algoliasearchHelper.url.getUnrecognizedParametersInQueryString(qs);
```
vvo: `getExtraParametersInQueryString`?

test/spec/helper.state.js
```diff
@@ -361,7 +361,7 @@ test('getConfigurationFromQueryString should parse page as number and be consist
 
   var queryString = helper.getStateAsQueryString({filters: filters});
 
-  var partialStateFromQueryString = algoliasearchHelper.AlgoliaSearchHelper.getConfigurationFromQueryString(
```
vvo: Do we have a check that ensure that the previous `algoliasearchHelper.AlgoliaSearchHelper.` methods are still there and works?

bobylito: no

README.md
bobylito: Not bad. @redox did a little research yesterday on the topic and it seems that unrecognized should be pretty common.

vvo: Ok I was not sure myself but if unrecognized is native-compliant then allright. Unrecognized gives more detail about what we are searching for, `extra` is generic indeed.

test/spec/helper.state.js
vvo: Maybe we can add one so that no developer will remove it (BUT when we do a V3 for example).

vvo: Maybe we can add one so that no developer will remove it (BUT when we do a V3 for example).

bobylito:
Ok will add some compat tests :)

bobylito:
I added some compat tests and updated the doc. We're good to go, @vvo ?

test/spec/compat.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+var test = require('tape');
+var algoliaSearchHelper = require('../../index');
+var url = algoliaSearchHelper.url;
+var AlgoliaSearchHelper = algoliaSearchHelper.AlgoliaSearchHelper;
+
+test('Should be compatible', function(t) {
+  t.equal(
+    AlgoliaSearchHelper.getConfigurationFromQueryString,
```
vvo: ahah nice trick :)

vvo:
:fireworks: go

PR #232, opened 2015-10-14
fix(numerics) : removeNumericFilters(attribute) was not working FIX #230

### LINKED ISSUE #230: removeNumericRefinement(attribute)
Hi there,

[According to the README.md](https://github.com/algolia/algoliasearch-helper-js/blob/master/README.md#batch-numeric-filter-removal) we should be able to do `removeNumericRefinement(attribute)`

But if you look at the code, it looks like you need to specify at least the couple `(attribute, operator)`: https://github.com/algolia/algoliasearch-helper-js/blob/43bae7b06366effa13328500bf4bb59682157316/src/SearchParameters/index.js#L700-L732

Did you remove this functionality?

bobylito:


bobylito:
Anyone up for a review @redox @pixelastic @pixelastic @proudlygeek ?

src/SearchParameters/index.js
```diff
@@ -705,16 +705,16 @@ SearchParameters.prototype = {
...
     if (paramValue !== undefined) {
+      if (!this.isNumericRefined(attribute, operator, paramValue)) return this;
       return this.setQueryParameters({
         page: 0,
         numericRefinements: this._clearNumericRefinements(function(value, key) {
           return key === attribute && value.op === operator && value.val === paramValue;
         })
       });
-    } else if (operator) {
+    } else if (operator !== undefined) {
```
proudlygeek: This way you lose checks against ```null```and ```false```. Is what you want?

p.s. Even if it's slower I usually use ```typeof operator === 'undefined'```, which has the advantage of working also when the variable is not defined.

bobylito: Yes you're right! And I do want to check the number of parameters, I do not care much about their type. It's also done for consistency with existing test, a few lines above.

proudlygeek:
LGTM :+1:

bobylito:
Cool thanks for the review @proudlygeek :)

src/SearchParameters/index.js
pixelastic: You might even remove the `else` here, because you always `return` in the previous `if`

src/SearchParameters/index.js
```diff
@@ -1139,12 +1140,15 @@ SearchParameters.prototype = {
...
-   * @param {string} operator operator of the refinement
+   * @param {string} [operator] operator of the refinement
    * @param {string} [value] value of the refinement
    * @return {boolean} true if it is refined
    */
   isNumericRefined: function isNumericRefined(attribute, operator, value) {
     if (isUndefined(value)) {
+      return !!this.numericRefinements[attribute];
+    }
+    if (isUndefined(value)) {
```
pixelastic: Hmm, this check will never pass, because you already checked for that and returned above

bobylito: that's a bummer :/

test/spec/helper.numericFilters.js
```diff
@@ -80,6 +80,52 @@ test('Should be able to remove values one by one even 0s', function(t) {
...
+
+  t.deepEqual(
+    helper.getRefinements('attribute'),
+    [{type: 'numeric', operator: '<', value: [4]}],
+    'should have the same result with getRefinements');
+
+  t.end();
+});
+
+test('Should remove all the numeric values for a single operator if remove is called with two arguments', function(t) {
```
pixelastic: Bad copy paste :)

test/spec/helper.numericFilters.js
```diff
@@ -80,6 +80,52 @@ test('Should be able to remove values one by one even 0s', function(t) {
...
+
+  t.deepEqual(
+    helper.getRefinements('attribute'),
+    [],
+    'should have the same result with getRefinements');
+
+  t.end();
+});
+
 test('should be able to retrieve numeric filters', function(t) {
```
pixelastic: Empty test?

src/SearchParameters/index.js
bobylito: I don't agree on the remove everything that is not mandatory policy ;)

test/spec/helper.numericFilters.js
bobylito: Indeed :/

src/SearchParameters/index.js
pixelastic: It just adds more branches (`if`, `else`, etc) which makes code harder to read. The less code you write, the less bug you have :)

bobylito: I do not agree, it expresses the completness of the tests without having to read inside :)

pixelastic: Yep, makes sense. More of personal preference, without much incidence I guess.

bobylito: That is something we agree on :D

PR #234, opened 2015-10-15
feat(SearchParameters): toggleRefinement on SearchParameters

bobylito:
Quick refactoring to move the auto magic toggleRefinement method from the helper to the state so that it can be used in state generation (like in the case of url generation)


redox:
LGTM :+1:

PR #241, opened 2015-10-26
chore: upgrade deps, shrinkwrap everything, lint, remove jsfmt

vvo:
+ upgrade all deps
+ remove jsfmt (not seen where it brings value vs eslint, eg: instantsearch dev)
+ lint
+ shrinkwrap the whole
+ upgrade browzers, test on microsoft edge

As done in all our "critical" js libraries (helper, client, instant).


vvo:
IE8 tests are broken. Helper is no more IE8 compatible. We need to indicate to use polyfill.io now

PR #244, opened 2015-11-02
fix(hierarchical): refined + no result

maxiloc:


maxiloc:
@vvo Not sure how to test this one

vvo:
Can you elaborate on what you are fixing with maybe a list of actions leading to the bug (this will help build the test case)

maxiloc:
If you refine the hierarchical menu and then type a query that lead to no results.
https://github.com/algolia/instantsearch.js/issues/385

vvo:
So you could start with simple.js again and make it so you have two actions (first toggle where you have results, then do a new search where there are no results).

Or did you already tried so?

maxiloc:
Yes that's ok, I am just not sure what to assert

vvo:
You should assert on the output of getFacetValues I guess?

maxiloc:
@vvo I think it's good to go

vvo:
:clap:

PR #246, opened 2015-11-03
Fix #245 : hitsPerPage not parsed as a number

### LINKED ISSUE #245: HitsPerPage attribute not parsed as a number
Linked to https://github.com/algolia/instantsearch.js/issues/366

bobylito:
This PR also contains : 
  - adds missing attribute minProximity in SearchParameters
  - updates links (were outdated)
  - adds other missing parameters
  - adds missing documentation for filter on SearchParameters

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -206,6 +206,11 @@ function SearchParameters(newParameters) {
    */
   this.minWordSizefor2Typos = params.minWordSizefor2Typos;
   /**
+   * Configure the precision of the proximity ranking criterion
+   * @see https://www.algolia.com/doc/rest#param-minProximity
+   */
+  this.minProximity = params.minProximity;
+  /**
    * Should the engine allow typos on numerics.
    * @see https://www.algolia.com/doc/rest#param-allowTyposOnNumericTokens
    * @member {boolean}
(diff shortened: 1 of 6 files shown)


src/SearchParameters/index.js
```diff
@@ -358,8 +363,8 @@ function SearchParameters(newParameters) {
     if (!this.hasOwnProperty(paramName)) {
       // IE8/9 has no console (BUT if devtools opened), nevermind there's no
       // developer working ONLY in IE8/9
-      window.console &&
-      window.console.error(
+      console &&
```
redox: why replacing `window.console` by `console`? That code won't work on old IE.

bobylito: Because it rely on window which breaks which unit testing with node. I will improve that.

bobylito:
Tests are still broken. Weird thing is that it's on MS Edge about 400 answer from the server...

vvo:
If tests are I need to work a little on the test stack indeed. Uf ecerything seems fine and tests ok in node chrome and change is a simple one, merge

bobylito:
I wouldn't go as far as to say it is trivial but it's pretty simple so I guess it's good to merge :) (and the failing test is already in previously merged and released work :P )

bobylito:
Fixes #245

vvo:
Ah yes that's just IE8 failing, the helper is no more compatible. Once we release instantsearch I will need to work a lil bit on this.

vvo:
Can you think of any application relying on hitsPerPage being a string that we would break?

bobylito:
https://www.algolia.com/doc/rest#param-hitsPerPage it is an integer. So it will break one way or another, if it is not parseable

vvo:
LGTM

PR #248, opened 2015-11-11
fix(hierarchicalWidget): Error when faceted and no results

maxiloc:
This one occurs when no results with query + refinements and the second query with query without refinement returns result.


maxiloc:
fixes https://github.com/algolia/instantsearch.js/issues/532

bobylito:
LGTM :) Thanks :)

PR #257, opened 2015-11-26
fix(core): makes node dependencies explicit FIX #256

### LINKED ISSUE #256: Makes the helper compatible with React-Native

bobylito:


redox:
:+1: LGTM

PR #259, opened 2015-11-27
fix(request-builder): Makes queries less ambiguous for client fix #205

### LINKED ISSUE #205: Query parameter sent twice in the query params
- on http://demos.algolia.com/instant-search-demo/
- looking at the network tab
- then clicking on a queries/* call
- then headers
- then scrolling to form data, clicking on view source
 - you will see this:

2015-09-08-114759_772x79_scrot

Which shows the double query=&query= parameter. So it means we send it twice in the parameters.

bobylito:
Query parameter was sent twice in the query params.

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -19,7 +19,6 @@ var requestBuilder = {
     // One query for the hits
     queries.push({
       indexName: index,
-      query: state.query,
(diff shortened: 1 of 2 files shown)


bobylito:
@vvo @pixelastic can you check this one?

vvo:
WFM!

PR #260, opened 2015-11-27
Adds getNumericRefinement to the helper (fix #214)

### LINKED ISSUE #214: expose state.getNumericRefinement(attribute, operator) as helper.getNumericRefinement
When reading the numeric methods on the helper you get:
- helper.addNumericRefinement(attribute, operator, value)
- helper.removeNumericRefinement(attribute, operator, value)

But no helper.getNumericRefinement(attribute, operator).

Do you think it would make sense to expose it? We have state.getNumericRefinement but it's not obvious because other methods are on the helper instance.

bobylito:


vvo:
LGTM

Issue #262, opened 2015-12-04
Expose optionalTagFilters and optionalFacetFilters


redox:
Those new (non-documented yet) parameters are not supported by the helper yet; I think we can move forward and add them.

PR #263, opened 2015-12-08
Feat/optionalTagFilters & optionalFacetFilters

olance:
Just a basic support for raw parameters as a quick fix to #262.

Should we add more support for those?
I'm thinking:

* `clearOptionalTags` to mimic `clearTags`
* `clearOptionalFacets` maybe?
* `addOptionalFacet`/`addOptionalFacets` or something similar to be able to pass an array of facet filters?

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -159,9 +159,7 @@ function SearchParameters(newParameters) {
   this.tagFilters = params.tagFilters;
 
   /**
-   * Contains the tag filters in the raw format of the Algolia API. Setting this
-   * parameter is not compatible with the of the add/remove/toggle methods of the
-   * tag api.
+   * Contains the  optional tag filters in the raw format of the Algolia API.
    * @private
    * @see https://www.algolia.com/doc/rest#param-tagFilters
    * @member {string}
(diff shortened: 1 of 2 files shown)


vvo:
Good for a first step, @olance I think we should now aim at solving the full PR like you proposed: provide the necessary functions to add, get, clear optional tags and facets.

@bobylito Do you think it makes sense to add those functions?

bobylito:
That's cool. Do we need the API on the helper right away? Do we mix optional and non-optional tags / filters?

vvo:
> Do we mix optional and non-optional tags / filters?

I would say no at first.

If optional tag/filters are used in different UI elements than the non optional ones then we should separate the methods and not mix them. cc @olance How are they used?

> Do we need the API on the helper right away?

Why not, if it's not so big to do let's do it now

bobylito:
Well I think that's good enough for now. I'm creating an issue to add proper support #267

olance:
Sorry @vvo I hadn't seen your comment earlier.
This should indeed probably be kept separate as optional tag/facet filtering and tag/facet filtering are really addressing different, and complementary use cases.

For instance, tag/facet filtering to search products present in a single shop + optional tag filtering to push up products bought by the current user.

PR #264, opened 2015-12-08
fix(search-results): adds missing results parameters Fix #261

### LINKED ISSUE #261: Expose aroundLatLng in results
When `aroundLatLngViaIP` is used in the query, please expose latitude and longitude in the results. It looks like this `aroundLatLng` is returned by the algolia API but the helper does not expose it.

Intended use case is displaying approximate distance between the IP address and the individual results.

bobylito:


vvo:
LGTM

PR #265, opened 2015-12-10
Fix/hierarchical facets order

olance:
When using hierarchical facets, attributes composing the hierarchy tree are declared in a specific order:

```js
var helper = algoliasearchHelper(client, indexName, {
    hierarchicalFacets: [{
      name: 'categories',
      attributes: ['categories.lvl0', 'categories.lvl1']
    }]
});
```

It can happen (at least it happened to me!) that the API will return facets data for those attributes in a different order than the expected one:

```json
{
    "results": [{
      ... 
      "facets": {
        "categories.lvl1": { "beers > IPA": 6, "beers > 1664": 3 },
        "categories.lvl0": { "beers": 9 }
      }
    }, 
    ...
  ]
};
```
Note that lvl1 comes *before* lvl0 here!

The current code iterates on each key of `facets`, pushing each facet data to the `SearchResults.hierarchicalFacets` array in the order of the keys... which will later silently break the generation of the hierarchical facets tree!

This PR uses the declared attributes order for the hierarchical facet to make sure the facets data is correctly ordered.

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -277,11 +277,14 @@ function SearchResults(state, algoliaResponse) {
     var hierarchicalFacet = findMatchingHierarchicalFacetFromAttributeName(state.hierarchicalFacets, facetKey);
 
     if (hierarchicalFacet) {
-      this.hierarchicalFacets[findIndex(state.hierarchicalFacets, {name: hierarchicalFacet.name})].push({
+      // Place the hierarchicalFacet data at the correct index depending on the attributes order that was defined at the
+      // helper initialization
+      var facetIndex = hierarchicalFacet.attributes.indexOf(facetKey);
+      this.hierarchicalFacets[findIndex(state.hierarchicalFacets, {name: hierarchicalFacet.name})][facetIndex] = {
         attribute: facetKey,
         data: facetValueObject,
         exhaustive: mainSubResponse.exhaustiveFacetsCount
-      });
+      };
     } else {
       var isFacetDisjunctive = indexOf(state.disjunctiveFacets, facetKey) !== -1;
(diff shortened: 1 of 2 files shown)


olance:
hmm test failed because:

> <microsoftedge 20 on Windows 10> [INT][FILTERS] Should retrieve different values for multi facetted records
> Error: AlgoliaSearchError: Cannot connect to the AlgoliaSearch API. Send an email to > support@algolia.com to report and resolve the issue. Application id was: test

:/

redox:
Good catch @olance :+1: @vvo @bobylito what do you think?

vvo:
Will launch the test locally, should be good

vvo:
About the API facet order that's really weird. Since the begining we have taken into account that the result order of a batch query was the same than the request order

vvo:
OK I think I get it and this is a good fix, just investigatin a bit more to understand it fully

vvo:
LGTM

To sum up, the API server sends back facets like this:

```js
var facets = {
facetName: [],
facetName2: []
}
```

And we are using a lodash.forEach to loop against `facets` and then mapping the loop index to the attributes: ['..', '..'] array

But that was not good since you cannot guarantee the iteration of a map in JavaScript.

So what made @olance is first try to map the facetName to the attributes: ['', ''] array and then fill the results with it.

This can have consequences like having holes in the results so he also used compact to solve it.

This is a fix that works but in the next helper iteration we should try to build the results from the top (facet definitions in the helper) rather than the bottom (what the API sent back).

So that we do not have to use things like compact etc..

bobylito:
Good! Thanks @olance :)

olance:
:)
Happy to help!

PR #266, opened 2015-12-10
test(CI): IE8 now requires a polyfill

vvo:
IE8 is dying, but the helper can still work if you include a polyfill.

Like this:

```html
<!--[if lte IE 8]>
<script src="https://cdn.jsdelivr.net/core-
js/2.0.0-alpha/shim.min.js"></script>
<![endif]-->
```

FTR I did not use https://cdn.polyfill.io/v2/docs/ because of a weird
"out of stack space" bug that I traced down to the Object.create
polyfill ONLY when running the tests (crazy browserify stuff
hapening).

I would say core-js seems to be working nicer and including it inside
a conditionnal comment should solve our issues.

I am looking forward simplifying our testing stack, it's currently
heavy with lot of code we do not need to be added (like tape
browserify streams etc..)

Also upgrade browserify, watchify, zuul and browzers (list of reusable browsers to test)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1298,3 +1298,15 @@ Here is an example of a result object you get with the `result` event.
    "index": "bestbuy"
 }
 ```
+
+# Browser support
+
+This project works fine on any [ES5](https://en.wikipedia.org/wiki/ECMAScript#5th_Edition) browser, basically >= IE9+.
+
+To get IE8 support you will have to include this:
+
+<!--[if lte IE 8]>
+  <script src="//cdn.jsdelivr.net/core-js/2/shim.min.js"></script>
+<![endif]-->
+
+Which will simulate most of the ES5 features in IE8.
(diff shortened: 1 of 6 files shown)


vvo:
weeee passing PR!

vvo:
Do not merge right now

test/run.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-require('bulk-require')(__dirname, ['spec/**/*.js']);
+// require('bulk-require')(__dirname, ['spec/**/*.js']);
```
bobylito: Maybe not necessary to keep that? :)

bobylito:
Looks good :) Thanks for taking the time to do that!

test/run.js
vvo: Just doing a test to see whats wrong in microsoft edge (working locally though), will remove then

vvo:
Ok edge is failing only because its an old version on saucelabs it seems, works in real life

PR #278, opened 2016-02-12
fix(pagination): adds doc on reset behavior, changes the name of setter

bobylito:
The `helper.setCurrentPage` is deprecated in favor of `helper.setPage`

### DIFF
--- a/README.md
+++ b/README.md
@@ -642,14 +642,37 @@ helper.clearTags().search();
 #### Get the current page
 
 ```js
-helper.getCurrentPage();
+helper.getPage();
 ```
 
 #### Change page
 
 ```js
-helper.setCurrentPage(3).search();
+helper.setPage(3).search();
 ```
+
+#### Automatic reset to page 0
+
+During a search, changing the parameters will update the result set, which can then change
+the number of pages in the result set. Therefore, the behavior has been standardized so
+that any operation that may change the number of page will reset the pagination to page 0.
+
+This may lead to some unexpected behavior. For example:
+
+```js
+helper.setPage(4);
+helper.getPage(); // 4
+helper.setQuery('foo');
+helper.getPage(); // 0
+```
+
+Non exhaustive list of operations that trigger a reset:
+ - refinements (conjunctive, exclude, disjunctive, hierarchical, numeric)
+ - tags 
+ - index (setIndex)
+ - setQuery
+ - setHitsPerPage
+ - setTypoTolerance
 
 ### Index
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -651,6 +651,30 @@ helper.getCurrentPage();
...
+that any operation that may change the number of page will reset the pagination to page 0.
+
+This may lead to some unexpected behavior. For example:
+
+```js
+// helper.state.page = 0
+helper.setPage(4);
+// helper.state.page = 4
+helper.setQuery('foo');
+// helper.state.page = 0
```
vvo: I would not talk about helper.state, just:

```js
helper.setPage(4);
helper.setQuery('foo');
helper.search(); // page will be 0, even if 4 was asked
```

bobylito: Fair enough!

README.md
```diff
@@ -651,6 +651,30 @@ helper.getCurrentPage();
...
+```
+
+Non exhaustive list of operations that trigger a reset :
```
vvo: Non exhaustive because there are two many of them? Maybe we can do the opposite list (or it's not needed?)

vvo: `reset :` no space before ` :`

README.md
vvo: And maybe say that you need to do the setPage last if you really really want page 4

src/algoliasearch.helper.js
```diff
@@ -366,22 +366,33 @@ AlgoliaSearchHelper.prototype.nextPage = function() {
  * @return {AlgoliaSearchHelper}
  */
 AlgoliaSearchHelper.prototype.previousPage = function() {
-  return this.setCurrentPage(this.state.page - 1);
+  return this.setPage(this.state.page - 1);
 };
 
+function setCurrentPage(page) {
+  if (page < 0) throw new Error('Page requested below 0.');
```
vvo: Do we need to test this?

README.md
bobylito: What about using getPage (well getCurrentPage for now :P, will fix that)?

vvo: In the example? I would say we only need to explain the trap people will fall into but I understand the concern about also explaining that the internal state is modified (for power users)

src/algoliasearch.helper.js
bobylito: Well that's the current behavior and the idea behind when implemented was "best do it before than later"

bobylito:
I think we're good here. Thanks for the feedback @vvo

PR #279, opened 2016-02-12
Test(distinct): tests behavior of facets using distinct

### LINKED ISSUE #272: Do not apply distinct for the second query of the disjunctive facets.
For example my dataset contains tshirts and shoes and they all have colors. I apply a distinct on the colors, I would like to choose tshirts and see all the colors. I can't right now because the facet values only contains those in the result set (which is limited to a single color).

Would it make sense to not apply distinct to the secondary queries? Should it be an option?

cc/ @redox @vvo @pixelastic @kokliKo

bobylito:
This test is intended to test that if we are using distinct, the facet
that we retrieve are not filtered using distinct. FIX #272 

So basically, this adds a test without actually changing the API because the behavior is the one expected.

### DIFF
--- a/test/integration-spec/helper.distinct.facet.js
+++ b/test/integration-spec/helper.distinct.facet.js
@@ -1,0 +1,71 @@
+'use strict';
+
+var utils = require('../integration-utils.js');
+var setup = utils.setup;
+
+var algoliasearchHelper = utils.isCIBrowser ? window.algoliasearchHelper : require('../../');
+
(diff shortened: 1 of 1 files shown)


vvo:
Awesome, took me some time to fully understand it but makes sense.

Do you think it might trigger any bad behavior for already in production people?

bobylito:
No I'm just testing the issue #272 is not an issue ;)

PR #280, opened 2016-02-15
feat(shortener): public with a set method

Jerskouille:
I've highlighted the reason for this PR there: https://github.com/algolia/instantsearch.js/issues/838 .  
Here is the example provided:

> - Zendesk's `/search` page accepts a `query` parameter.
> - I want to keep the default back-end search (which uses `query`) to have decent SEO on a query
> - However, by activating instantsearch, I end up with two parameters in the url `/search?query=a&q=a`
> - When I change the state the url becomes `/search?query=a&q=abc`.
>   If someone shares this URL, the back-end search will return results for `a` and the front-end results for `abc`, which is pretty uncool.
> 
> If I could change the "shortened" version of `query` from `q` to `query`, everything would be ultra simple.

New API:
```js
import algoliasearchHelper from 'algoliasearch-helper';
algoliasearchHelper.url.shortener.set('query', 'queryParam');
// `query` is  now globally mapped to `queryParam`
```

### DIFF
--- a/src/SearchParameters/shortener.js
+++ b/src/SearchParameters/shortener.js
@@ -1,5 +1,11 @@
 'use strict';
 
+/**
+ * Module handling the shortened versions of the query
+ * string parameters
+ * @module algoliasearchHelper.url.shortener
+ */
+
 var invert = require('lodash/object/invert');
 var keys = require('lodash/object/keys');
 
@@ -59,13 +65,14 @@ var keys2Short = {
 };
 
 var short2Keys = invert(keys2Short);
+var ENCODED_PARAMETERS = keys(short2Keys);
 
 module.exports = {
   /**
    * All the keys of the state, encoded.
    * @const
    */
-  ENCODED_PARAMETERS: keys(short2Keys),
+  ENCODED_PARAMETERS: ENCODED_PARAMETERS,
   /**
    * Decode a shorten attribute
    * @param {string} shortKey the shorten attribute
@@ -81,5 +88,22 @@ module.exports = {
    */
   encode: function(key) {
     return keys2Short[key];
+  },
+  /**
+   * Set an attribute / short attribute pair
+   * @param {string} key the attribute
+   * @param {string} shortKey the shorten attribute
+   */
+  set: function(key, shortKey) {
+    var oldShort = keys2Short[key];
+    var idx = ENCODED_PARAMETERS.indexOf(oldShort);
+    if (idx === -1) {
+      ENCODED_PARAMETERS.push(shortKey);
+    } else {
+      ENCODED_PARAMETERS[idx] = shortKey;
+      delete short2Keys[oldShort];
+    }
+    keys2Short[key] = shortKey;
+    short2Keys[shortKey] = key;
   }
 };
--- a/src/url.js
+++ b/src/url.js
@@ -35,17 +35,22 @@ function recursiveEncode(input) {
   return input;
 }
 
-var refinementsParameters = ['dFR', 'fR', 'nR', 'hFR', 'tR'];
 var stateKeys = shortener.ENCODED_PARAMETERS;
-function sortQueryStringValues(prefixRegexp, a, b) {
+function sortQueryStringValues(
+  prefixRegexp,
+  queryParameter,
+  refinementsParameters,
+  a,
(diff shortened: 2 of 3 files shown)


Jerskouille:
I have no clue on why the build failed. `npm run test` passes on my machine and I can't find any KO in the log on travis.

src/SearchParameters/shortener.js
```diff
@@ -81,5 +88,22 @@ module.exports = {
...
+   * @param {string} shortKey the shorten attribute
+   */
+  set: function(key, shortKey) {
+    var oldShort = keys2Short[key];
+    var idx = ENCODED_PARAMETERS.indexOf(oldShort);
+    if (idx === -1) {
+      ENCODED_PARAMETERS.push(shortKey);
+    } else {
+      ENCODED_PARAMETERS[idx] = shortKey;
+      delete short2Keys[oldShort];
```
bobylito: Isn't `oldshort` a value? Also you can't/shouldn't delete an array entry with delete. You should use splice but I don't think it is necessary, since you are replacing the value.

vvo:
> I have no clue on why the build failed. npm run test passes on my machine and I can't find any KO in the log on travis.

Yes sorry about that this is an issue of the build process

src/SearchParameters/shortener.js
```diff
@@ -59,13 +65,14 @@ var keys2Short = {
...
 var short2Keys = invert(keys2Short);
+var ENCODED_PARAMETERS = keys(short2Keys);
 
 module.exports = {
   /**
    * All the keys of the state, encoded.
    * @const
    */
-  ENCODED_PARAMETERS: keys(short2Keys),
+  ENCODED_PARAMETERS: ENCODED_PARAMETERS,
```
vvo: Why this change, to reflect that it's a constant?

vvo: Sorry, got it this is to reuse it later on

src/SearchParameters/shortener.js
vvo: Is it needed to delete the old short => key relation?

bobylito: consistency?

vvo:
About the API entry, what do you think on another take:

Use a mapping object:
```js
getStateAsQueryString({
  q: 'query',
  hFR: 'hCat'
});
```

And this mapping object can be passed to any url related stuff.

That way you do not have to learn or get the long tokens of every refinement, you only have to read the url and say "No, I do not want `q`, I want `query`"?

bobylito:
> That way you do not have to learn or get the long tokens of every refinement, you only have to read the url and say "No, I do not want q, I want query"?

Right it is not practical and the shortener is not very well documented, agreed. But I'm not sure about adding your proposition (which is clearly interesting) because it is not a main stream feature... Maybe it could help to have this kind of runtime override, instead of relying on a state and it could be easier to grasp.

src/SearchParameters/shortener.js
Jerskouille: Exactly @bobylito , here we replace it in `keys2Short`, so there's no reason it stays in `short2Keys` which is supposed to be `invert(keys2Short)`.

Jerskouille:
> Along with the PR, it would be cool to inline in the PR description the underlying API users would have to use (helper.url.shortener.set). We do have new tests (YAY!) but I was not able to quickly determine the underlying new API we would want to expose.

Agreed, added.

Jerskouille:
About using the mapping directly, I agree, something I'm not found of in my solution is that I'm overriding a global. I'll hack up another PR so that we can compare.

bobylito:
> I'll hack up another PR so that we can compare.

Do you want to go forward with this PR then?

Jerskouille:
Well, the other PR will include more changes (hence will be more prone to failure) + will decrease performance of the URL methods which might be a tradeoff we don't want to afford, so I'd say let's keep it open for now, and when I push the other one pick between the two.

bobylito:
Ok awesome :)

Jerskouille:
You can now find an untested version here: https://github.com/algolia/algoliasearch-helper-js/pull/281 for comparison.

vvo:
closing this one

PR #281, opened 2016-02-15
feat(url): new mapping option for URL methods

Jerskouille:
To compare with #280 .

Here the new API consists of adding a `mapping` option to the URL methods.
The mapping would be between the current shortened version and the new one, like so:

```js
{
  q: 'query',
  hFR: 'hCat'
}
```

The methods which accept this new `mapping` option are:
```js
algoliasearchHelper.url.getStateFromQueryString(qs, {mapping: {hFR: 'hCat'}})
algoliasearchHelper.url.getUnrecognizedParametersInQueryString(qs, {mapping: {hFR: 'hCat'}})
algoliasearchHelper.url.getQueryStringFromState(state, {mapping: {hFR: 'hCat'}})
```

### DIFF
--- a/src/url.js
+++ b/src/url.js
@@ -20,6 +20,7 @@ var mapValues = require('lodash/object/mapValues');
 var isString = require('lodash/lang/isString');
 var isPlainObject = require('lodash/lang/isPlainObject');
 var isArray = require('lodash/lang/isArray');
+var invert = require('lodash/object/invert');
 var encode = require('qs/lib/utils').encode;
 
 function recursiveEncode(input) {
@@ -37,11 +38,14 @@ function recursiveEncode(input) {
 
 var refinementsParameters = ['dFR', 'fR', 'nR', 'hFR', 'tR'];
 var stateKeys = shortener.ENCODED_PARAMETERS;
-function sortQueryStringValues(prefixRegexp, a, b) {
+function sortQueryStringValues(prefixRegexp, invertedMapping, a, b) {
   if (prefixRegexp !== null) {
     a = a.replace(prefixRegexp, '');
     b = b.replace(prefixRegexp, '');
   }
+
+  a = invertedMapping[a] || a;
+  b = invertedMapping[b] || b;
 
   if (stateKeys.indexOf(a) !== -1 || stateKeys.indexOf(b) !== -1) {
     if (a === 'q') return -1;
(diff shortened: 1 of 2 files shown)


vvo:
aside from the internal changes, as a user, would you prefer using this API or the previous one? (#280)

Jerskouille:
I think this one makes more sense, especially if we abstract this away in `instantsearch` through a `urlSync.mapping` option.

src/url.js
```diff
@@ -77,10 +84,9 @@ exports.getStateFromQueryString = function(queryString, options) {
     partialStateWithPrefix,
     function(v, k) {
       if (prefixForParameters && prefixRegexp.test(k)) {
-        var encodedKey = k.replace(prefixRegexp, '');
-        return shortener.decode(encodedKey);
+        k = k.replace(prefixRegexp, '');
```
bobylito: What about not mutating the parameter? :)

```js
var key = prefixForParameters && prefixRegexp.test(k) ? k.replace(prefixRegexp, '') : k;
var decodedKey = shortener.decode(new2Old[key] || key);
```

bobylito:
I would have expected this PR to be more troublesome. Looks fine this way!

src/url.js
Jerskouille: I think you might prefer the new commit.

src/url.js
```diff
@@ -37,12 +38,15 @@ function recursiveEncode(input) {
 
 var refinementsParameters = ['dFR', 'fR', 'nR', 'hFR', 'tR'];
 var stateKeys = shortener.ENCODED_PARAMETERS;
-function sortQueryStringValues(prefixRegexp, a, b) {
+function sortQueryStringValues(prefixRegexp, new2Old, a, b) {
```
vvo: first thing I notice if coming to the code is "what is new?" "what is old?"

what about "userMapping" or "customMapping"?

vvo:
Happy too because there's even less implementation code I believe, or equal

src/url.js
Jerskouille: I agree these namings are not really explicit. Working on this.

Jerskouille:
New version with more explicit naming and tests added.

bobylito:
:+1:

bobylito:
screenshot 2016-02-16 16 38 09

Jerskouille:
Thanks for this, fixed in latest version.

vvo:
[LGTM](http://lgtm.in/i/ker1)

bobylito:
Thanks @Jerskouille

vvo:
awesome PR

PR #284, opened 2016-02-17
fix(numeric filters): Makes it possible to add then remove a string based numeric filter FIX #285

### LINKED ISSUE #285: Inconsistency between add and remove NumericRefinement
Follow-up of https://github.com/algolia/instantsearch.js/pull/847 . In the helper:
- `addNumericRefinement(attribute, operator, value)` type-checks value and convert it to a float if it's a string
- but `removeNumericRefinement(attribute, operator, value)` just does a simple `===` comparison

Which means that:
```js
var val = /* ... */;
helper.addNumericRefinement('attribute', '=', val);
helper.removeNumericRefinement('attribute', '=', val);
```
is actually not guaranteed to work, which is, to me, weird.

bobylito:


Jerskouille:
:+1: I think we should also handle the Array that you can pass to `addNumericRefinement`.

vvo:
> I think we should also handle the Array that you can pass to addNumericRefinement.

Yes, otherwise LGTM

bobylito:
@vvo, @Jerskouille can you take a little bit of time to re-review the changes :)?

src/SearchParameters/index.js
```diff
@@ -21,11 +21,18 @@ var pluck = require('lodash/collection/pluck');
 var defaults = require('lodash/object/defaults');
 var merge = require('lodash/object/merge');
 var deepFreeze = require('../functions/deepFreeze');
+var valToNumber = require('../functions/valToNumber');
 
 var filterState = require('./filterState');
 
 var RefinementList = require('./RefinementList');
 
+function findArray(vs, v1) {
```
vvo: A bit hard to quickly grasp the function signature, we have three variables named vs, v1 and v. Can you make them explicit?

Jerskouille:
LGTM !

bobylito:
Mergin!

Jerskouille:
:100: !

PR #286, opened 2016-02-17
fix(SearchParameters)Supports for unknown parameters

### LINKED ISSUE #200: Throw on unsupported parameter
We have been getting issues on new parameters being used and not yet available on the helper.

Right now we were silently failing making it hard to understand what's going on, following #198 we should throw in V3. Why V3? Because this could now possibly break a good number of implementations and websites.

BTW @bobylito can you re explain why do we need to declare all of the parameters explicitly instead of doing `forEach(params, param => this.paramName = this.paramValue)`?

bobylito:
Will warn once on unknown parameters.
Fixes #200 but instead of throwing, it will accept all parameters, just warn.

### DIFF
--- a/src/functions/warnOnce.js
+++ b/src/functions/warnOnce.js
@@ -1,0 +1,23 @@
+'use strict';
+var bind = require('lodash/function/bind');
+
+try {
+  var warn;
+
+  if (typeof window !== 'undefined') warn = window.console && bind(window.console.warn, console);
+  else warn = bind(console.warn, console); // eslint-disable-line no-console
+
+  var warnOnce = (function(w) {
+    var previousMessages = [];
+    return function warnOnlyOnce(m) {
+      if (previousMessages.indexOf(m) === -1) {
+        w(m);
+        previousMessages.push(m);
+      }
+    };
+  })(warn);
+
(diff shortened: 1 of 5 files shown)


src/functions/warnOnce.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+var bind = require('lodash/function/bind');
+
+var warn = window ?
+  window.console && bind(window.console.warn, console) :
+  bind(console.warn, console); // eslint-disable-line no-console
+
+var warnOnce = (function(w) {
+  var previousMessages = [];
```
vvo: The array will only grow by the number of different parameters we use right?

vvo:
LGTM

src/functions/warnOnce.js
bobylito: Yes, it will grow by one for each new message generated. Do you think it might a problem? Maybe we should deactivate that in the minified version?

vvo: Should be ok since it only grows on the number of "unknown" params, not by the number of calls to setQueryParameter etc..

bobylito:
The war (with the tests) is over :) Merging!

PR #287, opened 2016-02-19
fix(numeric filters): Makes it possible to add then remove a string based numeric filter

bobylito:
It also fixes the usage of arrays: we can now remove arrays from numeric filters.


bobylito:
Same as #284 but on the correct branch :)

PR #293, opened 2016-03-15
Perf refactoring of SearchParameters._parseNumbers

pixelastic:
I had a perf issue on instantsearch. When enabling the `urlSync`, typing in the input was really slow and sluggish. After a bit of digging, it turns out that the issue was coming from the fact that I had to fetch more than 1000 facet values and that the widget was generating a unique url for each of them.

After looking at the Chrome CPU profiling, I found that most of the time was spent in `SearchParameters._parseNumbers` in the helper.
image

I started by putting a few tests on the method, than starting refactoring. Mostly, I used the lodash `cloneDeep` method instead of `merge`, `mapValues` instead of a custom-made loop and exported a bit of logic into a recursive method.

My empirical tests showed that it was faster, but I also added a performance benchmark in `./test/perf/SearchParameters/_parseNumbers.js` that you can run with `npm run test:perf`. Here is the output:

```
Scores: (bigger is better)

parseNumbers with cloneDeep
Raw:
 > 87.47514910536779
 > 87.56218905472637
 > 80.1186943620178
 > 86.73978065802592
Average (mean) 85.47395329503448

parseNumbers with merge
Raw:
 > 38.272816486751715
 > 38.767395626242546
 > 38.72889771598808
 > 38.423645320197046
Average (mean) 38.54818878729485

Winner: parseNumbers with cloneDeep
Compared with next highest (parseNumbers with merge), it's:
54.9% faster
2.22 times as fast
0.35 order(s) of magnitude faster
QUITE A BIT FASTER
```

So this refactor is actually a bit faster. Not sure this will fix my instantsearch issue, but this is a good start.

I'm also unsure about keeping the perf benchmark in the repo. On one hand I find it interesting to prove that it is actually faster than before, on the other hand I'm not sure it is useful to keep it in the repo like this. Tell me what you think @bobylito @vvo and I'll remove it from the PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "test": "scripts/test.sh",
     "test-node": "clear && node test/run",
     "test:watch": "onchange '{src,test}/**/*.js' 'index.js' -- npm run test-node",
+    "test:perf": "node ./test/perf/**/*.js",
     "release": "git clean dist/ -f && git checkout dist/ && npm install && ./scripts/release.js && npm run build",
     "shrinkwrap": "npm-shrinkwrap --dev"
   },
@@ -49,6 +50,7 @@
   },
   "devDependencies": {
     "algoliasearch": "latest",
+    "bench": "^0.3.6",
     "browserify": "^12.0.1",
     "browzers": "^1.3.0",
     "bulk-require": "^0.2.1",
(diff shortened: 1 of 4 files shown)


pixelastic:
Wrong branch

PR #294, opened 2016-03-15
Performance refactor of SearchParameters._parseNumbers

pixelastic:
I had a perf issue on instantsearch. When enabling the `urlSync`, typing in the input was really slow and sluggish. After a bit of digging, it turns out that the issue was coming from the fact that I had to fetch more than 1000 facet values and that the widget was generating a unique url for each of them.

After looking at the Chrome CPU profiling, I found that most of the time was spent in `SearchParameters._parseNumbers` in the helper.
image

I started by putting a few tests on the method, than starting refactoring. Mostly, I used the lodash `cloneDeep` method instead of `merge`, `mapValues` instead of a custom-made loop and exported a bit of logic into a recursive method.

My empirical tests showed that it was faster, but I also added a performance benchmark in `./test/perf/SearchParameters/_parseNumbers.js` that you can run with `npm run test:perf`. Here is the output:

```
Scores: (bigger is better)

parseNumbers with cloneDeep
Raw:
 > 87.47514910536779
 > 87.56218905472637
 > 80.1186943620178
 > 86.73978065802592
Average (mean) 85.47395329503448

parseNumbers with merge
Raw:
 > 38.272816486751715
 > 38.767395626242546
 > 38.72889771598808
 > 38.423645320197046
Average (mean) 38.54818878729485

Winner: parseNumbers with cloneDeep
Compared with next highest (parseNumbers with merge), it's:
54.9% faster
2.22 times as fast
0.35 order(s) of magnitude faster
QUITE A BIT FASTER
```

So this refactor is actually a bit faster. Not sure this will fix my instantsearch issue, but this is a good start.

I'm also unsure about keeping the perf benchmark in the repo. On one hand I find it interesting to prove that it is actually faster than before, on the other hand I'm not sure it is useful to keep it in the repo like this. Tell me what you think @bobylito @vvo and I'll remove it from the PR.

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -1,10 +1,13 @@
-'use strict';
 
+'use strict';
 var keys = require('lodash/object/keys');
 var intersection = require('lodash/array/intersection');
 var forOwn = require('lodash/object/forOwn');
 var forEach = require('lodash/collection/forEach');
 var filter = require('lodash/collection/filter');
+var cloneDeep = require('lodash/lang/cloneDeep');
+var mapValues = require('lodash/object/mapValues');
+var includes = require('lodash/collection/includes');
 var map = require('lodash/collection/map');
 var reduce = require('lodash/collection/reduce');
 var omit = require('lodash/object/omit');
@@ -443,7 +446,7 @@ SearchParameters.PARAMETERS = keys(new SearchParameters());
  * @return {object} a new object with the number keys as number
  */
 SearchParameters._parseNumbers = function(partialState) {
-  var numbers = {};
+  var newState = cloneDeep(partialState);
 
   var numberKeys = [
     'aroundPrecision',
@@ -459,36 +462,33 @@ SearchParameters._parseNumbers = function(partialState) {
     'minProximity'
   ];
 
-  forEach(numberKeys, function(k) {
-    var value = partialState[k];
-    if (isString(value)) numbers[k] = parseFloat(partialState[k]);
+  // parse numeric root keys to floats
+  newState = mapValues(newState, function(value, key) {
+    if (!includes(numberKeys, key) || !isString(value)) {
+      return value;
+    }
+    return parseFloat(value);
   });
 
-  if (partialState.numericRefinements) {
-    var numericRefinements = {};
-    forEach(partialState.numericRefinements, function(operators, attribute) {
-      numericRefinements[attribute] = {};
-      forEach(operators, function(values, operator) {
-        var parsedValues = map(values, function(v) {
-          if (isArray(v)) {
-            return map(v, function(vPrime) {
-              if (isString(vPrime)) {
-                return parseFloat(vPrime);
-              }
-              return vPrime;
-            });
-          } else if (isString(v)) {
-            return parseFloat(v);
-          }
-          return v;
-        });
-        numericRefinements[attribute][operator] = parsedValues;
-      });
-    });
-    numbers.numericRefinements = numericRefinements;
+
+  // Parse input to float, act recursively on arrays
+  function recurseParseFloat(input) {
+    if (isArray(input)) {
+      return map(input, recurseParseFloat);
+    }
+    return parseFloat(input);
   }
 
-  return merge({}, partialState, numbers);
+  var numericRefinements = newState.numericRefinements;
+  if (numericRefinements) {
+    numericRefinements = mapValues(numericRefinements, function(refinements) {
+      return mapValues(refinements, recurseParseFloat);
+    });
+    newState.numericRefinements = numericRefinements;
+  }
+
+
+  return newState;
 };
 
 /**
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -54,6 +54,10 @@
         }
       }
     },
+    "bench": {
+      "version": "0.3.6",
+      "resolved": "https://registry.npmjs.org/bench/-/bench-0.3.6.tgz"
+    },
     "browserify": {
       "version": "12.0.1",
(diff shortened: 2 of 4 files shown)


pixelastic:
I'm having Travis errors with iPhone here. I don't really know how to fix them...

vvo:
Awesome that we are spending some time on the helper performance because yes in instantsearch.js the top slow functions are coming from the helper.

On parseNumbers side, if I understand the code, we call it at every method call of the helper => every time we generate a new state. Since we generate a new state at each method call (immutability) then parseNumbers is called at every method call.

@bobylito Is it necessary to call parseNumbers every time? I would think we only need to parse numbers when parsing a stringified state => when creating a state from a url string.

I am sure I miss some context to fully understand it, let us know!

vvo:
> I'm having Travis errors with iPhone here. I don't really know how to fix them...

"No big deal" (as in "it's expected to fail", it's been some time since I need to fix the multi browser env)

pixelastic:
I did a quick bench of removing the call to parseNumbers to every SearchParameter creation and yes, it improves a lot:

```
noParseNumbers
Raw:
 > 622.754491017964
 > 622.3776223776224
 > 622.3776223776224
 > 624.750499001996
Average (mean) 623.0650586938012

base
Raw:
 > 82.66932270916335
 > 82.50497017892644
 > 82.83433133732535
 > 82.91708291708292
Average (mean) 82.7314267856245

Winner: noParseNumbers
Compared with the slowest (base), it's:
86.72% faster
7.53 times as fast
0.88 order(s) of magnitude faster
```

I will try to see how we could limit the number of times this is called. From what I understand it is used to make sure every possible numeric value in the API options is actually a number and not a string. It checks a list of known keys, as well as all numericRefinements.

I'll see first how the API reacts to passing strings instead of numbers. Maybe it is a problem we do not need to solve because the API already handles it.

Then I'll try to add numeric conversion to numericRefinements only when we add them, instead of on each creation.

WDYT @bobylito?

src/SearchParameters/index.js
```diff
@@ -459,36 +462,33 @@ SearchParameters._parseNumbers = function(partialState) {
     'minProximity'
   ];
 
-  forEach(numberKeys, function(k) {
-    var value = partialState[k];
-    if (isString(value)) numbers[k] = parseFloat(partialState[k]);
+  // parse numeric root keys to floats
+  newState = mapValues(newState, function(value, key) {
```
bobylito: if you're looking for optimization, you should not loop over the keys of the newstate but on the numberKeys (fewer number of iterations). Also mapValues will create a new object. Best would be to do: 

```js
forEach(numberKeys, function(k) {
  var value = newState[k];
  if (isString(value)) newState[k] = parseFloat(newState[k]);
});
```

bobylito:
The code was actually checking if the value was a string. So it's more conservative than just parsefloat everything :)

bobylito:
Maybe we could just have a string based API and in this case force every input to string. But it makes the API way less useful... Maybe we could provide methods that return numbers, and keep only the string internally. That could work.

src/SearchParameters/index.js
pixelastic: I wasn't sure which one had the larger number of keys. I'll follow your advice.

pixelastic:
I don't think we should force anything to a specific type at such a top level. I would simply accept what the user is giving us blindly, except for a few specific cases (like when decoding the url, or setting numeric refinements).

bobylito:
@pixelastic It's not enough. To be able to compare two states we need to have the same types when we let the user do something to the data, or if we deserialize it.

> except for a few specific cases (like when decoding the url, or setting numeric refinements)

what are the other cases?

pixelastic:
When do we need to compare states?

I don't have many more examples in mind. Mostly I think we should validate (or automatically cast values) on the initial input (for example in instantsearch, not in the helper).

vvo:
> When do we need to compare states?

More precisely when would a hitsPerPage comes as a string or a number? From a url => string, but we can just parse number when deserialazing the url maybe?

bobylito:
> But then we can just parse number when deserialazing the url maybe?

@vvo: But isn't that the problem @pixelastic found? He tried to remove all the parsing and his urlSync problems were gone...

pixelastic:
Well, not exactly. Optimizing _parseNumbers helped, and completly removing it helped even better, but I have other perf issues (not all of them coming from the helper).

But now I'm confused. If we completly remove _parseNumbers (and instead precisely cast to float specific values at input), would the state still be immutable? I'm asking because we are doing a deepClone in _parseNumbers, so if I remove _parseNumbers, I still need to create a new fresh copy of the state, right?

src/SearchParameters/index.js
pixelastic: Actually I tried, and I tried your suggestion of trying to return early without cloning if we detect that there is nothing to do, and it was even slower :/

pixelastic:
Ok so let's focus on this single code change for the moment. I tried to make it faster (just focusing on the inside of the _parseNumbers methods), and did not manage any significant improvements. All my tests actually made it slower.

I think it is still an improvement on what it was before (even if not a major one), and could be merged like this.

Let's split the other possible improvements into other issues maybe?

src/SearchParameters/index.js
bobylito: :D

bobylito:
Well for now it's good. But before any further perf improvement, we should have reproducible perf tests.

Shall we merge? @pixelastic @vvo

pixelastic:
I think we can close this issue.

My suggested solution with a `deepClone` instead of a merge, even if faster, is still not the best solution. As discussed in #299, it is better to use an `assign` here.

I've created #297 that only contains the `_parseNumbers` unit tests.

PR #297, opened 2016-03-17
test(_parseNumbers): Add tests for _parseNumbers

pixelastic:
Adding tests to _parseNumbers. We are going to use assign instead of
the deepClone anyway, so I removed the previous parseNumbers
suggestion.


vvo:
LGTM, moar tests moar fun

PR #303, opened 2016-03-21
perf(deepFreeze): Removing the calls to deepFreeze

### LINKED ISSUE #301: Avoid deepFreeze
We currently deepFreeze the resulting new state generated by the helper as an extra security to avoid people messing up with the state properties directly instead of using the helper methods.

This is pretty expensive because it's again a deep traversal + Object.freeze execution on each node of the searchParameters instance on each state modification.

We can avoid doing so because nowhere we document modifying properties of the searchParameters directly: we always document using methods like setQuery.

## TODO

*(modify this list when needed if you start working on it)*

- remove deepFreezing of the final searchParameters instance

pixelastic:
Fixes #301

As discussed in #301, we currently do not really need to call
deepFreeze on the new state. The operation itself is expensive and we
do not really need this protetion currently because we always document
modifying the object through its methods.


vvo:
LGTM

PR #314, opened 2016-06-09
Fix(_parseNumber): Store unparseable strings for root query parameter

### LINKED ISSUE #313: AroundRadius now supports 'all' as a value
But as a convenience, this parameter was parsed as a number... And now it creates a problem because 'all' is not parseable. This problem could happen with any kind of those number based query parameters.

Proposed behaviour:
 - if the new value is a number, keep the number
 - still parse numbers if it's a string
 - if it fails to parse the value, then store the string (but no fail)

bobylito:
Some number parameters can also have string values. Storing number as numbers ensures that the same doesn't have 2 representations. If it's not parseable it should still be stored.

Fixes #313


vvo:
LGTM, new test, passing tests at least unit ones on nodejs

PR #315, opened 2016-06-10
chore(deps): upgrade all deps, try at fixing CI

vvo:


vvo:
Hey that passed! So tests should no morebe flaky.

PR #317, opened 2016-06-20
feat(urls): provide a `safe` option to fully encode the url

vvo:
following algolia/instantsearch.js#982,
I propose to add a `safe` parameter to getQueryStringFromState that
allows you to retrieve the fully encoded url instead of a custom
encoded non-safe one.

### DIFF
--- a/src/url.js
+++ b/src/url.js
@@ -134,17 +134,22 @@ exports.getUnrecognizedParametersInQueryString = function(queryString, options) 
  *  - mapping : map short attributes to another value e.g. {q: 'query'}
  *  - moreAttributes : more values to be added in the query string. Those values
  *    won't be prefixed.
+ *  - safe : get safe urls for use in emails, chat apps or any application auto linking urls.
+ *  All parameters and values will be encoded in a way that it's safe to share them.
+ *  Default to false for legacy reasons ()
  * @return {string} the query string
  */
 exports.getQueryStringFromState = function(state, options) {
   var moreAttributes = options && options.moreAttributes;
   var prefixForParameters = options && options.prefix || '';
   var mapping = options && options.mapping || {};
+  var safe = options && options.safe || false;
   var invertedMapping = invert(mapping);
-  var partialStateWithEncodedValues = recursiveEncode(state);
+
+  var stateForUrl = safe ? state : recursiveEncode(state);
 
   var encodedState = mapKeys(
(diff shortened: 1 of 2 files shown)


src/url.js
```diff
@@ -134,17 +134,24 @@ exports.getUnrecognizedParametersInQueryString = function(queryString, options)
...
+  var safe = options && options.safe || false;
   var invertedMapping = invert(mapping);
-  var partialStateWithEncodedValues = recursiveEncode(state);
+
+  if (safe === false) {
+    state = recursiveEncode(state);
+  }
 
   var encodedState = mapKeys(
-    partialStateWithEncodedValues,
```
bobylito: As of coding style in the helper, we try to avoid reusing variables and especially parameters.

vvo: I am fine with that rule we should enforce then

bobylito: 👍 :)

vvo:
modified code and added a simple test

vvo:
adding another test for safe

vvo:
Ok ready for review :)

src/url.js
```diff
@@ -134,17 +134,27 @@ exports.getUnrecognizedParametersInQueryString = function(queryString, options)
...
  * @return {string} the query string
  */
 exports.getQueryStringFromState = function(state, options) {
   var moreAttributes = options && options.moreAttributes;
   var prefixForParameters = options && options.prefix || '';
   var mapping = options && options.mapping || {};
+  var safe = options && options.safe || false;
   var invertedMapping = invert(mapping);
-  var partialStateWithEncodedValues = recursiveEncode(state);
+  var stateForUrl;
```
bobylito: What about:

```javascript
var stateForUrl = options.safe ? state : recursiveEncode(state);
```

And also drop l.146?

vvo: I need var safe (for later usage and to follow in place conventions about options here), but agreed on the first part.

bobylito:
Reviewing!

See comments :) Nice new tests 👍

src/url.js
bobylito: Yeah ok makes sense!

bobylito:
All green!!! Thanks for your patience @vvo 👯

vvo:
Weeeeeeeeeeeeeeeeeeeeee

PR #319, opened 2016-06-23
Added an AngularJS example in the main README

seafoox:


bobylito:
Thanks for the contribution @seafoox

PR #324, opened 2016-07-12
feat(middlewares): add middlewares

Morhaus:
In the React implementation of Instantsearch.js, we pass a helper to all our connected components thanks to `algoliasearch-helper-provider`. However, some components need to register a default/initial configuration. For now, in order to isolate this default configuration, I'm creating another helper and monkey-patching search/searchOnce and proxying events (https://github.com/algolia/instantsearch.js/blob/cbf89e5b43cb8bd4ea8b80267280ee793c5cf5a2/packages/instantsearch-react/src/createConfigManager.js#L57).

The reason why this configuration need to be isolated is so that we can differentiate it from search parameters that have been set as the result of an action from the user (change hitsPerPage/refine facet/etc.) and that should be persisted to the URL.

Having middlewares built-in to the helper makes this much cleaner to implement.

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -41,6 +41,7 @@ function AlgoliaSearchHelper(client, index, options) {
   this.lastResults = null;
   this._queryId = 0;
   this._lastQueryIdReceived = -1;
+  this._middlewares = [];
 }
 
 util.inherits(AlgoliaSearchHelper, events.EventEmitter);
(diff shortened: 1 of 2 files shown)


vvo:
While I get the implementation and API you are adding, I am unable to see exactly WHY we need it. And why it needs to be implemented inside the helper and not userland in instantsearch-react.

bobylito:
That's a really great feature! I think this look really helpful.

However, I'm a bit uneasy with the semantic and purpose. I understand that in your case, it's very useful and you know the purpose and everything. But out of context, the concept of middleware doesn't appear very obvious. Where does it apply? Can it apply on other steps of the lifecycle? Do we have other use cases? Can the feature that you need be integrated directly in the helper?

PR #325, opened 2016-07-15
New documentation

bobylito:
instantsearch.js relies internally on the helper and when users want to implement new widgets, they need to use and somehow understands the Helper. With the impulse that we have to create more variations of instantsearch.js for frameworks (the React one is in progress), a good documentation is most needed.

This PR proposes a completely new documentation. Previously we had the automatically generated documentation and some outdated examples. This new version proposes the following enhancements:

 - new introductions to the helper: a true getting started strictly focused on using the Helper, and a higher level view presenting the concepts
 - a semi automatically generated documentation. Like for instantsearch.js using the jsDoc for content but restructured for users.
 - up to date examples that document comment patterns and usage with JS frameworks.

The big focus points are exhaustivity, learning curve and readability.

Even though this is a PR to replace the documentation, it is a work in progress and as we move forward we'll improve. Now we'll have the fondations to easily do that.

This documentation [live on my own fork of the library](http://bobylito.me/algoliasearch-helper-js/documentation/).


rayrutjes:
Here are my 2cts

1. This hurts my eyes
image

2. Can we maybe augment the size a bit in the navigation?
image

Looks great!!! Keep it up!

redox:
> This hurts my eyes

Indeed, I think we could just stick to our pink (#fb366e) color instead of that green one.

> Can we maybe augment the size a bit in the navigation?

And maybe also add some left padding to sub-sections.

bobylito:
> Indeed, I think we could just stick to our pink (#fb366e) color instead of that green one.

I wanted to differentiate this documentation from the places website. The colors will be changed.

LukyVj:
DESKTOP : 

You miss the active state on the navigation links, like this on places : 

-

The header will stay "grey" ? 

It feels "unachieved" to me.
-

Missing wording ? 

-

You could probably emphasize the color background here : 

-

Give those links the same color and aspect as the others ? 

-

I have the feeling that thin light borders would do the trick here, to show how the data is structured : 

-

There is also a sidebar bug on the examples page : 

-

About the codepen embeds, not sure if you need them, but if you want to use the "light" theme, and show only the result panel, use `data-theme-id="41"`


MOBILE : 
You miss the "nav to select" component and the icon on mobile : 


2 Bugs on mobile navigation : 
The color should be white ? 
It disapear after a bit of scroll



-- 

Other than that, it's really cool! So far so good! :) Keep it up! 🤘

pixelastic:
image
Missing title

image
I think using the "instant search" naming here will be confusing in regard to instantsearch.js. We need to clearly explain what the helper library is about, and we should avoid "instant search" IMHO.

Agreed on the green contrast that's close to unreadable to me :)

Nice diagrams in the core concepts page

The Reference page is huuuuuge. Might be interesting to have either a small TOC of each method under each title in the main area, or a sub-list in the sidebar.

Maybe tone done the color of the return value (because almost all methods return a `AlgoliaSearchHelper`) and its bold and black, while the method name is gray and normal.

I like the examples with each methods. We could improve them by actually using the passed argumenst (`error`, `state`, `lastResults`) instead of simply `console.log`. This would show what to do with those variables.

This was just a shallow pass, I did not read all the content yet.

bobylito:
> I think using the "instant search" naming here will be confusing in regard to instantsearch.js. We need to clearly explain what the helper library is about, and we should avoid "instant search" IMHO.

Happy to get some alternatives. The main page is definitely not the best part. @pixelastic 

> Nice diagrams in the core concepts page

Thanks 


> The Reference page is huuuuuge. Might be interesting to have either a small TOC of each method under each title in the main area, or a sub-list in the sidebar.

How can do a sub navigation? especially with black header that stays :( @Shipow @AntoinePlu 

> I like the examples with each methods. We could improve them by actually using the passed argumenst (error, state, lastResults) instead of simply console.log. This would show what to do with those variables.

Love the idea. Contributions welcome :)

bobylito:
> Maybe tone done the color of the return value (because almost all methods return a AlgoliaSearchHelper) and its bold and black, while the method name is gray and normal.

Do you have a screenshot of that? I'm not sure to see what you're talking about @pixelastic

vvo:
This is a great achievement to have this website almost finished! I like the fact that you are introducing a simple color scheme so that we immediately know on which community website we are.

As I reported to you earlier, I am really missing a true method reference page instead of a navigation + ctrl+f your method page like you did.

The helper is currently made of a good number of methods so having them visually at all time would help me navigate the reference API.

Plus the fact that everytime I click a "see the API reference" link, I am really expecting a method list. I do not have the perfect menu list to use but I am frustrated a bit by the current one. I will still need to ctrl+f the page.

bobylito:
> As I reported to you earlier, I am really missing a true method reference page instead of a navigation + ctrl+f your method page like you did.

Your point of view is very valuable and I think it would add value to be able to have it. Nonetheless, the vision is to first bring on board people who are not familiar with the API and then help those who can have an idea of which method to select.

If you're a newcomer it helps to have all the methods organised by topic instead of having them organised alphabetically. As you said there are a lot of methods, and some works with certain features but not others. For example, if I want to work with tags, it helps to go see the methods for the tags. Also this hierarchy gives a highlight of the available features.

Let's keep your suggestion in mind for now, and let's ship it like that. If it doesn't work, then we'll fix it.

vvo:
> Let's keep your suggestion in mind for now, and let's ship it like that. If it doesn't work, then we'll fix it.

WFM!

pixelastic:
@bobylito: Here is what I was referring to about the title / black / gray / bold
image
`setPage` is the important thing here while `SearchParameters` is secondary (most of the methods are actually returning a `SearchParameters`). I would have pu more emphasis on the method name and less on the return type.

---

I'm now reading the whole doc and putting my comments as they come. I'll add rewording suggestions as well.

---

We'll need a favicon (easier to spot the correct tab in my tablist).

---

Not sure about the naming in the header. `algoliasearch-helper` seems too much like a github repo name. I suggest `algolia-helper.js`. But on the other hand, it __is__ the name of the repo... I still feel the project will need a fancier name.

---

I like the "toolbox" word to describe the helper. I would maybe remove all the `instant-search` instances with something like `awesome search`, `state-of-the-art search`, etc. For example:
- _The toolbox for creating instant-search experiences_ => The toolbox for creating state-of-the-art search experiences
- _Build instant-search experiences your way!_ => Build awesome search experiences your way!
- _Algolia Helper provides rock solid foundations for making your own instantsearch UX with your tools._ => Algolia Helper provides rock solid foundations for making your own search UX.

---

> The helper proposes a way to structure your search, so you can spent time on your users search

If you want to say that the Helper has its own structure (hooks, events, in a specific order), I would say this a bit differently. _proposes a way_ sounds like it's a default behavior that can be overriden (while it's not, actually). I don't yet know how to say it differently, but we need something that explain that there is some structure in the way the helper is built, and core concepts to understand.

---

> Parameters tracking - the helper is the search source of truth
> Event based - read and write operations can be decoupled
> High level features - easy access to advanced facetting techniques and more
> And more - smart behaviours, chainability...

Suggestion to make it more impactful (and why this lib is interesting, what issues does it solves)
Parameters tracking - single source of truth
Event based - only listen to what you care about
High level features - complex facetted features in a an easy to write syntax
And more - sane defaults, immutability, chainability

---

> The helper is any-framework friendly

=> Framework agnostic
The helper can be used as-is, or embedded into your framework of choice. See our examples with React, Backbone and Ember.

---

image
The three links here are broken, but I guess it will be fixed when we deploy on the real repo.

image
The links are using the `utm` tags of Places.

---

image
Not sure this link is needed. It only jumps one paragraph ahead.

---

I would rewrite a bit the part where we explain that the user can use either our own dataset or their own. We are saying the same thing multiple times and the fact that it's either one or the other could be made clearer. I will send a PR for the wording later, I'm only on the "big stuff" on this first read.

---

> As the helper extends the client, therefore it also needs to be included in your project.

`it` is ambiguous.

---

image
On my browser (`Version 48.0.2564.82 Ubuntu 15.04 (64-bit)`) I cannot scroll the horizontal scrollbar.

---

Throughout the _getting started_, I would rename the `index` variable to `index_name`. `index` is used in the SDK documentation but refers to a full object, not just a string. Clearly defining that we need to pass the index name and not an index object might make things easier to grasp.

---

image
We can add a link on _custom ranking_ to the documentation here.

---

image
The _Download this example_ link is not working

---

image
Is there a way to display the codepen in a more dynamic way? Here we can only see the results by default. It would be more entertaining to see the code and being able to interact with it directly instead of having to click on HTML/JS/CSS. What would be perfect is displaying three panels (JS/HTML/results) directly.

---

image
Very nice recap. Makes learning the concepts easier as we go step by step.

---

> We're going to focus on the actual results/records computed by Algolia.

We should select one naming (result, record, object, item, whatever) and stick to it throughout the whole website. This is actually a bigger issue than just this website because we are using various names accross all our documentations, but let's try to stay consistent locally here.

---

image
This can be simplified like this:
```
$('#container').html(function() {
    return $.map(content.hits, function(hit) {
      return '<li>' + hit.name + '</li>'
    });
  });
```
(Same for the following code examples)

---

image

Remove the extraneous `the`.

---

image
I would a link to the guide on facetting, or the relevant part of the documentation that explains facetting in more details.

---

image
We should explain here that `getFacetValue` is a special method added on the returned `content` for ease of use. If there are other hidden methods like that, we should tell the reader about them (or link to the relevant part of the doc on the website)

---

image
Here we add `data` through `attr('data-foo')` and later we read it through `.data()`. This requires good knowledge of jQuery. I would instead write data using `.data()`.

---

image
I'm a bit worried about that example. Here we allow clicks on any part of the `#facet-list`, still we try to read the `data` of the click target, and only the `checkbox` have such an attribute. Wouldn't it be easier to limit the click on the actual interesting elements, by doing `$('#facet-list').on('click', 'input[type=checkbox]', ...)` ? Or even better, completly remove the label/checkbox markup and use simple span/div for the sake of the example?

---

>  Those values are generated by Algolia based on the rest of the search, meaning that it will only 
provide facet values that are meaningful for the current other parameters. [...] This kind of facets are called conjunctive facets but they are not the only kind of filtering that you can apply with the helper.

I think we should be more clear on the fact that all facets will be dynamically updated based on the current choices, and that conjunctive facets means that the more filters you choose, the narrower the results will be. These are two different important distinctions.

---

image
Same here, this might be too complex with not enough explanation. Is it really possible to do Hierarchical Facetting easily (it requires special formatting of data in the index, as well as custom code in the front-end, it might be outside of the helper scope). Also, what is "tag facetting" that cannot be done with conjunctive/disjunctive facetting?

---

image
I would put the disclaimer about the fact that instantsearch might be more suitable for more people more up in the page, not at the very end. Like one of the first things in the getting started, to explain that the SDK is low level, the helper is a toolbox, and instantsearch.js is a library including UX and UI guidance.

---

Ok, that's a long enough post for today, I'll read the other pages later :)
image

bobylito:
Thanks @pixelastic, making the changes

Shipow:
great review @pixelastic , there are some points which I'd like to add in a list that we could check at every project release

bobylito:
> Is there a way to display the codepen in a more dynamic way? Here we can only see the results by default. It would be more entertaining to see the code and being able to interact with it directly instead of having to click on HTML/JS/CSS. What would be perfect is displaying three panels (JS/HTML/results) directly.

Layout doesn't give us enough to make this comfortable. Let's try to find a better way :)

bobylito:
> We should explain here that getFacetValue is a special method added on the returned content for ease of use. If there are other hidden methods like that, we should tell the reader about them (or link to the relevant part of the doc on the website)

This method is documented. I will link it.

Shipow:
release the kraken! :)

Issue #327, opened 2016-07-25
Adding methods to add and remove facets from the facets attribute


Morhaus:
Right now the way to set and edit the facets attribute of the helper's state and of a SearchParameters is either through instantiation

```js
new SearchParameters({facets: ['genre']})
```

or through manipulating the facets attribute directly

```js
state.setFacets('facets', union(state.facets, ['genre']));
```

We should have methods to add and remove facets from the facets attribute.

* `state.addFacet(facet)`
* `state.removeFacet(facet)`
* `state.addDisjunctiveFacet(facet)`
* `state.removeDisjunctiveFacet(facet)`
* `state.addHierarchicalFacet(facet)`
* `state.removeHierarchicalFacet(facet)`

PR #328, opened 2016-07-26

Morhaus:
See #327

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -1020,7 +1020,7 @@ SearchParameters.prototype = {
       return this;
     }
 
-    return this.setQueryParameters({
+    return this.clearRefinements(facet).setQueryParameters({
       facets: filter(this.facets, function(f) {
         return f !== facet;
       })
@@ -1038,7 +1038,7 @@ SearchParameters.prototype = {
       return this;
     }
 
-    return this.setQueryParameters({
+    return this.clearRefinements(facet).setQueryParameters({
       disjunctiveFacets: filter(this.disjunctiveFacets, function(f) {
         return f !== facet;
       })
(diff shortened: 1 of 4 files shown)


src/SearchParameters/index.js
```diff
@@ -960,6 +1009,60 @@ SearchParameters.prototype = {
...
+   * @param {string} facet facet name to remove
+   * @return {SearchParameters}
+   */
+  removeFacet: function removeFacet(facet) {
+    if (!this.isConjunctiveFacet(facet)) {
+      return this;
+    }
+
+    const newFacets = this.facets.slice();
+    newFacets.splice(newFacets.indexOf(facet), 1);
```
bobylito: Why not use `filter` like for `removeTagRefinement`:

```js
      tagRefinements: filter(this.tagRefinements, function(t) { return t !== tag; })
```

src/SearchParameters/index.js
```diff
@@ -960,6 +1009,60 @@ SearchParameters.prototype = {
...
+   * @return {SearchParameters}
+   */
+  removeDisjunctiveFacet: function removeDisjunctiveFacet(facet) {
+    if (!this.isDisjunctiveFacet(facet)) {
+      return this;
+    }
+
+    const newFacets = this.disjunctiveFacets.slice();
+    newFacets.splice(newFacets.indexOf(facet), 1);
+    return this.setQueryParameters({
```
bobylito: What do we do with the already set refinements? Maybe we should clean them up?

src/SearchParameters/index.js
Morhaus: I'm used to avoiding using filter as a remove function, but filter does make for cleaner code. 👍

src/SearchParameters/index.js
Morhaus: Adding a refinement to a facet that is not present in `facets` throws, so maybe we should throw here too?

vvo: I would prefer as a user to not have to clear all refinements before remove a facet but that may be a bad shortcut?

bobylito:
Great PR :) Only problem is that it targets master :) Can you open a new one that targets develop?

PR #330, opened 2016-07-26

Morhaus:
See #327, #328


src/SearchParameters/index.js
```diff
@@ -17,6 +17,7 @@ var isUndefined = require('lodash/isUndefined');
 var isString = require('lodash/isString');
 var isFunction = require('lodash/isFunction');
 var find = require('lodash/find');
+var findIndex = require('lodash/findIndex');
```
bobylito: This is now unused :)

Morhaus: Wow, good catch! I need to make my ESLint warnings more obvious, I keep missing them.

PR #333, opened 2016-07-28

Morhaus:
Let's try this again. See #327, #328, #330, #332.

@bobylito @vvo we sometimes use `conjunctive` to describe the facets defined in the `facets` attribute. Should `{add,remove}Facet` be renamed to `{add,remove}ConjunctiveFacet`?

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -921,14 +921,15 @@ SearchParameters.prototype = {
   },
   /**
    * Add a hierarchical facet to the hierarchicalFacets attribute of the helper
-   * configuration, if it isn't already present.
+   * configuration.
    * @method
    * @param {object} hierarchicalFacet hierarchical facet to add
    * @return {SearchParameters}
+   * @throws will throw an error if a hierarchical facet with the same name was already declared
    */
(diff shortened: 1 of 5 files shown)


src/SearchParameters/index.js
```diff
@@ -960,6 +1009,60 @@ SearchParameters.prototype = {
...
+    }
+
+    return this.clearRefinements(facet).setQueryParameters({
+      disjunctiveFacets: filter(this.disjunctiveFacets, function(f) {
+        return f !== facet;
+      })
+    });
+  },
+  /**
+   * Add a hierarchical facet to the hierarchicalFacets attribute of the helper
```
bobylito: Shouldn't it be remove?

bobylito: Shouldn't it be remove?

bobylito:
> @bobylito @vvo we sometimes use conjunctive to describe the facets defined in the facets attribute. Should {add,remove}Facet be renamed to {add,remove}ConjunctiveFacet?

Definitely a fair question. But the attribute we use is facets, not conjunctiveFacets... Keeping it like that is okayish for now, imho.

PR #334, opened 2016-08-01
Fix footer img protocol

LukyVj:


bobylito:
Thanks @LukyVj for fixing this!

Issue #343, opened 2016-08-08
Make SearchParameters dumb and leave the magic in the main Helper.


bobylito:
Right now, we implement the automatic `setPage(0)` directly in the `SearchParameters` which is ok as long as we don't use the SearchParameters as a target of a reduce.

@Morhaus had the issue with the UI focused model made for instantsearch-react. In this context, it stores models (closer to the UI) that are reduced to a single SearchParameters that will be then used for generating a search. This pattern doesn't work if the order of the parameters are to be specific.

This would be a breaking change for those who use the search parameters directly AND expect the `setPage(0)` to happen.

bobylito:
fixed via #344

PR #344, opened 2016-08-08
feat(no-set-page): don't reset the page to 0 in SearchParameters methods

Morhaus:
See #343.

BREAKING CHANGE: SearchParameters methods don't reset the page to 0 anymore.


bobylito:
Looks good. Could you add a small something in the concept page part about the page behaviour?

PR #355, opened 2016-08-22
feat(helper) expose clearCache from the client Fix #345

### LINKED ISSUE #345: Expose the clearCache() method
Could be nice to expose the underlying `client.clearCache()` so we don't need to guess the `helper.client.clearCache()` method and we're not confused about the cache at the index-level (not used by the helper since we're doing multi-queries only).

```js
var helper = algoliasearchHelper(client, ....);
helper.clearCache();
```

bobylito:


bobylito:
Needs proper documentation.

vvo:
WFM

PR #359, opened 2016-08-23
test(getFacetValues) Adds unit tests for ordering

### LINKED ISSUE #354: What is the behaviour of getFacetValues without order for isRefined?
In the instantsearch documentation, we advise the use of isRefined without an order.

In the helper this translates to: 

```js
results.getFacetValues('city', 'isRefined');
```

Seems this behaviour might have changed recently...

bobylito:
+ cleaner test output
+ improve doc for getFacetValues

Fix #354


bobylito:
There is an actual regression of the behavior

vvo:
Isn't this a fix() more than just a test()?

Issue #250, opened 2015-11-16
{add,remove}HierarchicalFacetRefinement methods?


Jerskouille:
Here is what we currently have:

- Facet => `{add,remove}FacetRefinement`
- Disjunctive facet => `{add,remove}DisjunctiveFacetRefinement`
- Hierarchical facet => Only `toggleRefinement`
- Exclude => `{add,remove}FacetExclusion`
- Numeric => `{add,remove}NumericRefinement`
- Tag => `{add,remove}Tag`

We should probably add `addHierarchicalFacetRefinement` and `removeHierarchicalFacetRefinement` for consistency.

vvo:
Yes agree!

vvo:
Good for consistency but we should keep in mind that the hierarchicalFacet can only have one current refinement at a time, thus this is why I think add and remove where not implemented at first

bobylito:
I would like to be able to select a refinement wether or not it is already define. I think it's pretty close to add with the other kind of filters, but we can only select one value. Do you think it's close enough?

I would prefer that over checking if it's already refined.

I don't see any problems with the remove.

What do you think?

Jerskouille:
I think it makes sense @bobylito .

bobylito:
any thoughts @vvo ?

vvo:
so addHierarchicalRefinement would be an override? sort of?

vvo:
To sum up, yes we need addHierarchicalFacetRefinement and remove, add should throw if we are trying to add a second refinement.

There's only one possible refinement for every hierarchical facet at any point in time.

Anyone can take this.

PR #361, opened 2016-08-25
feat(hierarchicalFacets): add add and remove operations on hierarchical facets

bobylito:


redox:
Looks good :) Ping us when we need to review 👍

bobylito:
You can review @redox :)

src/SearchParameters/index.js
```diff
@@ -1280,6 +1280,41 @@ SearchParameters.prototype = {
...
+  /**
+   * Adds a refinement on a hierarchical facet.
+   * @param {string} facet the facet name
+   * @param {string} path the hierarchical facet path
+   * @return {SearchParameter} the new state
+   * @throws Error if the facet is not defined or if the facet is refined
+   */
+  addHierarchicalFacetRefinement: function(facet, path) {
+    if (this.isHierarchicalFacetRefined(facet)) {
+      throw new Error(facet + ' is already refined.');
```
redox: We never throw Errors in such case before. Would it make sense to let such double-refinement happen?

redox: But we may want to add a test to check if it's a hierarchical facet.

src/SearchParameters/index.js
```diff
@@ -1280,6 +1280,41 @@ SearchParameters.prototype = {
...
+  },
+
+  /**
+   * Removes the refinement set on a hierarchical facet.
+   * @param {string} facet the facet name
+   * @return {SearchParameter} the new state
+   * @throws Error if the facet is not defined or if the facet is not refined
+   */
+  removeHierarchicalFacetRefinement: function(facet) {
+    if (!this.isHierarchicalFacetRefined(facet)) {
```
redox: Check if it's a hierarchical facet first?

src/SearchParameters/index.js
bobylito: > But we may want to add a test to check if it's a hierarchical facet.

It's already done in `isHierarchicalFacetRefined`

> We never threw any errors in similar cases before. Would it make sense to let such double-refinement happen?

True. But it is how we ended up defining those methods in #250

src/SearchParameters/index.js
bobylito: See above comment :)

vvo:
LGTM

PR #363, opened 2016-08-26
fix(SearchParameters) Using unknown parameters not showing warning

### LINKED ISSUE #360: Remove warning with unknown parameters
More annoying and frustrating for users, than actually helpful.

bobylito:
This removes the warning that was output when an unknown parameter was
used. The engine will return an error if it does not understand a
parameter, therefore the `error` event should be checked carefully.

Fix #360


vvo:
LGTM, on loop less :)

PR #372, opened 2016-08-30
test(perf) add perf test

### LINKED ISSUE #298: Monitor the performance of the helper
To avoid regressions and be able to back up performance improvements we need to automatize performance measurements between commits.

The first step would be to create a test case where we do a regular helper usage like a user typing and refining stuff.

Then using https://github.com/isaacs/node-bench we should be able to precisely measure performance of such actions (time taken).

I believe search should not be called for such performance tests bcause I/O would mess up the measurements.

Just measuring CPU time on this should be sufficient for our needs.

Then once we have that setup we should setup on CI a script that will measure current version performance and commit modifications performance. So that we are always sure that there's no regression between current commit and current version.

The script should exit with 1 when thresold reached (something like +X% slower being the thresold)

## TODO

*(modify this list when needed if you start working on it)*

- [x] create a simple scenario of modifying the helper state (init the helper with params, setQuery, addRefinement...)
- [x] use node-bench to get performance measurements programmatically
- [x] setup the script that will compare current version and current commit
- [x] set the script to run on every PR

bobylito:
This provides the minimal archictecture for making perf assessment
tests. It will execute during each PR and merge on travis.

Fix #298

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ install: npm install
 script:
   - npm run test-ci
   - npm run doc
+  - npm run test:perf
 branches:
   only:
     - master
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "test": "scripts/test.sh",
     "test-node": "clear && node test/run",
     "test:watch": "onchange '{src,test}/**/*.js' 'index.js' -- npm run test-node",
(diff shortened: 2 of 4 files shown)


vvo:
Seems pretty cool!

PR #374, opened 2016-09-01
feat(derive) Let the user create a derivation of an helper

bobylito:
Make it possible to make multi index, parallel and aggregated queries.


test/integration-spec/helper.derive.js
```diff
@@ -0,0 +1,111 @@
...
+        t.equal(state.query, 'toto', 'The query `toto` should be used');
+        t.equal(results.hits.length, 1, 'Should retrieve one record');
+        t.equal(results.hits[0].objectID, '1', 'And it should be the record `1`');
+      });
+
+      helper.search();
+    }).then(null, bind(t.error, t));
+  });
+
+test(
```
PLNech: What is the difference between this test and the previous one?

test/spec/algoliasearch.helper/queryID.js
```diff
@@ -21,7 +21,7 @@ test('the queryid should keep increasing when new requests arrives', function(t)
   t.end();
 });
 
-test('the response handler should check that the query is not outdated', function(t) {
+test.skip('the response handler should check that the query is not outdated', function(t) {
```
PLNech: Why do you skip this test?

test/integration-spec/helper.derive.js
bobylito: Nice catch. Did I mention this is work in progress? :) I will add a more relevant second test :)

test/spec/algoliasearch.helper/queryID.js
bobylito: Because I reimplemented the function tested as an anonymous that makes use of closures so this is basically impossible to unit test right now :( If I move forward with this feature, I will make it testable again, and will reenable this test :)

vvo: Do we still have a test that checks it at a higher level maybe (Not getting too many "results" events given the outdated response).

bobylito: Can you further explain the test case you have in mind?

bobylito: I think that what you meant @vvo back then was "does it implement the outdated mechanism?" And the answer is yes (and the test is now available!!)

documentation-src/metalsmith/content/concepts.md
```diff
@@ -217,6 +217,27 @@ parameters object and its api to create new states that we can serialize.
 
 ![Search states](images/concepts/State is immutable.svg)
 
+### Derivations of the helper: multi-queries
```
vvo: Is multi index search one of the main usecases of this feature? Maybe we could name drop multi index somewhere then?

examples/readme.md
```diff
@@ -0,0 +1 @@
+The examples moved to [codepen](http://codepen.io/collection/nwWJbe/).
```
vvo: And also on the documentation

lezoudali: @vvo are there any good example on how to use this new feature? Can't find any examples related to this.

lezoudali:
@vvo are there any good example on how to use this new feature? Can't find any examples related to this.

Issue #399, opened 2016-10-05
When an error is returned from Algolia, a JS error occurs


kayneb:
When I have an invalid query, for example, if I specify an invalid index, I get a JS error:

```
index.js:210 Uncaught TypeError: Cannot read property 'results' of undefined
```

That's SearchResults/index.js.

The problem is that a `SearchResults` is being constructed with the Algolia response, which is undefined. That response is blindly accessed in the constructor of `SearchResults`.

bobylito:
Thanks for the issue! I looked at the PR, it is great :D

PR #400, opened 2016-10-05
#399: Passing null for results to callback when err present

kayneb:
There was an undefined error occurring when attempting to construct a `SearchResults` when there was no results (because of an error). This adds some falsey checking to ensure that results are present before constructing.

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -150,7 +150,8 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
     return this.client.search(
       queries,
       function(err, content) {
-        cb(err, !err ? new SearchResults(tempState, content) : null, tempState);
+        if (err) cb(err, null, tempState);
+        else cb(null, new SearchResults(tempState, content), tempState);
       });
   }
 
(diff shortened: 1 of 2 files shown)


kayneb:
This is for #399

src/algoliasearch.helper.js
```diff
@@ -150,7 +150,7 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
     return this.client.search(
       queries,
       function(err, content) {
-        cb(err, new SearchResults(tempState, content), tempState);
+        cb(err, !err ? new SearchResults(tempState, content) : null, tempState);
```
bobylito: Good catch! Could you use an explicit if on the line before? Maybe use if / else to show which case is which and dumb down the paramters? Something like:

```js
if(err) cb(err, null, tempState);
else cb(null, SearchResults(tempState, content), tempState);
```

kayneb: Sure!

kayneb: Addressed.

bobylito:
Perf test failed, it's alright :) Thanks for the contribution @kayneb

PR #401, opened 2016-10-07
Fix minor typos

rayrutjes:
Let me know if I should also update this in the dist folder, but I suppose you do that when you actually release.


bobylito:
Thanks for the fix @rayrutjes

PR #411, opened 2016-10-21
feat(SearchParameters): add option `startPage`

iam4x:
Hello ✌️ 

I could not find a way only using `url-sync.js` like @vvo specified in the original issue to enable this option, so I've added a new `SearchParameters` option to use like this:

``` jsx
const search = instantsearch({
  appId: 'xxxx',
  apiKey: 'xxxx',
  indexName: 'xxxx',
  searchParameters: {startPage: 1},
});
```

I'm opening the PR to know if I'm going to the right direction before writing the tests and update the doc.

linked to: https://github.com/algolia/instantsearch.js/issues/1444

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -259,6 +259,11 @@ function SearchParameters(newParameters) {
    */
   this.page = params.page || 0;
   /**
+   * The start page number
+   * @member {number}
+   */
+  this.startPage = params.startPage || 0;
+  /**
    * How the query should be treated by the search engine.
    * Possible values: prefixAll, prefixLast, prefixNone
    * @see https://www.algolia.com/doc/rest#param-queryType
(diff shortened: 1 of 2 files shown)


bobylito:
Thanks @iam4x, that's a very interesting feature indeed.

I'm not sure about the semantic. What about using `paginationOffset` or `paginationStart`?

More conceptually, I really like the feature but it challenges a bit what was the expected content of the SearchParameters so far. I will get back to you on that :)

iam4x:
Thank's for the reply @bobylito 

* I was having the same feeling about the naming and I like `paginationOffset` 👍
* About the `SearchParameters` options I don't know the whole context yet and I couldn't tell if it was good or evil.

bobylito:
Well after discussing with the team it seems that this might not need an option... Not clear about what we should provide to avoid the confusions:
 - new comers expect pages to start at 1
 - people who knows algolia, expect start at 0
 - current implementations start at 0

On hold for the moment. But I'll come back to it later. Thanks again @iam4x this is still much appreciated :)

bobylito:
Closing for now. Thanks again for the contribution @iam4x

PR #423, opened 2016-11-24
feat(searchForFacetValues): implement a new method to be able to search into facet values

bobylito:
# Search for facet values

## Description of the feature

This new method on the helper lets a user search into facet values. The function takes the attribute and the query as parameters, and it returns a promise of the content of the search. Because this method is implemented on the helper, it takes also the current context / state: all the facets and parameters will be sent with the query. Also the returned value is altered to add an `isRefined` attribute so that we know if the value is already selected.

_A note on facets_
If the searched facet is conjunctive, the filters that are already set on it will be sent - whereas if the facet is disjunctive, the filters on this facet won't be sent.

## Code example

Given that we the following dataset:

```js
[
  {objectID: '1', f: 'ba', f2: ['b']},
  {objectID: '2', f: 'ba', f2: ['c', 'x']},
  {objectID: '3', f: 'ba', f2: ['d']},
  {objectID: '4', f: 'bb', f2: ['b']},
  {objectID: '5', f: 'bb', f2: ['c', 'y']}
]
```

With the following configuration:
  - searchableAttributes: ['f', 'f2'],
  - attributesForFaceting: ['f', 'f2']

	
```js
var helper = AlgoliasearchHelper(client, 'index', {
  facets: ['f', 'f2']
});

helper.addFacetRefinement('f2', 'c');

helper.searchForFacetValues('f2', '').then(function(content) {
  console.log(content);
  /*
{
  facetHits: [
    {value: 'c', highlighted: 'c', count: 2, isRefined: true},
    {value: 'x', highlighted: 'x', count: 1, isRefined: false},
    {value: 'y', highlighted: 'y', count: 1, isRefined: false}
  ],
  processingTimeMS: 1
}
  */
});
```

⚠️ There is a test failt because of an issue when there is no filter at all. @Nagriar is on it :)

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -176,11 +176,11 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
 };
 
 /**
- * Search for a facet value based on an query and the name of the facetted attribute. This
+ * Search for facet values based on an query and the name of a facetted attribute. This
  * triggers a search and will retrun a promise. On top of using the query, it also sends
  * the parameters from the state so that the search is narrowed to only the possible values.
  * @param {string} query the string query for the search
- * @param {string} facet the name of the facetted attribute for which to search for a value
+ * @param {string} facet the name of the facetted attribute
  * @return {promise<FacetSearchResult>} the results of the search
  */
 AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query) {
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -176,6 +176,19 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
 };
 
 /**
+ * Search for a facet value based on an query and the name of the facetted attribute. This
+ * triggers a search and will retrun a promise. On top of using the query, it also sends
+ * the parameters from the state so that the search is narrowed to only the possible values.
+ * @param {string} query the string query for the search
+ * @param {string} facet the name of the facetted attribute for which to search for a value
+ * @return {promise<FacetSearchResult>} the results of the search
+ */
+AlgoliaSearchHelper.prototype.searchForFacetValues = function(query, facet) {
+  var index = this.client.initIndex(this.state.index);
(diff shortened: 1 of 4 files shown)


src/algoliasearch.helper.js
```diff
@@ -176,6 +176,28 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
...
+ * @param {string} query the string query for the search
+ * @param {string} facet the name of the facetted attribute for which to search for a value
+ * @return {promise<FacetSearchResult>} the results of the search
+ */
+AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query) {
+  var state = this.state;
+  var index = this.client.initIndex(this.state.index);
+  var isDisjunctive = state.isDisjunctiveFacet(facet);
+  var algoliaQuery = requestBuilder.getSearchForFacetQuery(facet, query, this.state);
+  return index.searchFacet(algoliaQuery).then(function addIsRefined(content) {
```
maxiloc: Shouldn't we rename the method in the client ?

vvo:
> Shouldn't we rename the method in the client ?

Yes

redox:
LGTM 👍

bobylito:
This PR is stalled until we merge https://github.com/algolia/algoliasearch-client-js/pull/362

vvo:
> This PR is stalled until we merge algolia/algoliasearch-client-js#362

Merged, released (3.19.2). Have fun :D

src/algoliasearch.helper.js
```diff
@@ -176,6 +176,28 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
 };
 
 /**
+ * Search for a facet value based on an query and the name of the facetted attribute. This
```
vvo: `for facet values`?

vvo:
Aside from small comment on doc. LGTM

src/algoliasearch.helper.js
bobylito: Humm it does not sound that right after all... Fixing that!

bobylito:
Merging even though there is a failing test. The fix will be released on the API next week.

PR #430, opened 2016-12-06
Fix(searchForFacetValues): Update the client method

bobylito:
Now using the new method name that has been implemented in the JS client.

Also the test still fail but a new engine is getting deployed and should be fixed soon.


bobylito:
Merging because it is only a naming fix.

test/integration-spec/helper.searchForFacetValues.js
```diff
@@ -79,7 +79,7 @@ test(
     }).then(null, bind(t.error, t));
   });
 
-test(
+test.only(
```
mthuret: don't forget to remove this ;)

PR #434, opened 2016-12-12
Removing tag filters from the documentation

### LINKED ISSUE #433: Remove tagFilters in the ReadMe
We're removing the notion of `tagFilters` in all the docs (deprecated).
Can we remove this paragraph from the ReadMe?

PS: the feature still needs to work, for legacy reasons.

kokliKo:
Fixes #433


kokliKo:
Ok, we're rolling back on this one, we don't have a consensus around how we should handle it.
I'll study the question in more details.

PR #448, opened 2016-12-19
Fix test for searchForFacetValues

bobylito:


bobylito:
Only perf failing :) This shouldn't be a blocker :(

PR #455, opened 2016-12-28
fix(agent): Add a test if addAlgoliaAgent exists

bobylito:
This was causing problems with users who are not upgrading their JS
Helper, but not their client.


redox:
I don't remember our rules about `console.log` but I've the feeling we should try to avoid them as much as possible. Here I assume it's not a big deal if we don't set the agent, isn't it @bobylito?

bobylito:
We do display some messages when they make sense. If the client they are using does not support the addAlgoliaAgent method, then it really is an old version.

PR #460, opened 2017-01-09
feat(client): Adds methods to set/get the client.

bobylito:
This is not really a feature for the users of the JS Helper, but more of
convenience for any library built on top of it.

See https://github.com/algolia/instantsearch.js/issues/1673

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -86,7 +86,7 @@ function AlgoliaSearchHelper(client, index, options) {
   if (!client.addAlgoliaAgent) console.log('Please upgrade to the newest version of the JS Client.'); // eslint-disable-line
   else client.addAlgoliaAgent('JS Helper ' + version);
 
-  this.client = client;
+  this.setClient(client);
   var opts = options || {};
   opts.index = index;
   this.state = SearchParameters.make(opts);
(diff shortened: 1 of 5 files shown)


src/algoliasearch.helper.js
```diff
@@ -1215,6 +1215,29 @@ AlgoliaSearchHelper.prototype.clearCache = function() {
...
 /**
+ * Updates the internal client instance. If the reference of the clients
+ * are equal then no update is actually done.
+ * @param  {AlgoliaSearch} newClient an AlgoliaSearch client
+ * @return {AlgoliaSearchHelper}
+ */
+AlgoliaSearchHelper.prototype.setClient = function(newClient) {
+  if (this.client === newClient) return this;
+
+  if (!newClient.addAlgoliaAgent) newClient.addAlgoliaAgent('JS Helper ' + version);
```
vvo: Should you call setClient from the constructor now that you have this init method?

test/integration-spec/helper.searchForFacetValues.js
```diff
@@ -25,8 +25,7 @@ var dataset = [
 ];
 
 var config = {
-  searchableAttributes: ['f', 'f2'],
-  attributesForFaceting: ['f', 'f2']
+  attributesForFaceting: ['searchable(f)', 'searchable(f2)']
```
vvo: Is this related to the client change?

bobylito: it's related to the engine changing its behavior...

bobylito: thus breaking the test...

vvo: Got it!

src/algoliasearch.helper.js
bobylito: Why not? I will try that!

bobylito: It works!

bobylito:
All changes done @vvo thanks for the review

PR #462, opened 2017-01-12
rename toggleRefinement to toggleFacetRefinement

### LINKED ISSUE #447: Rename toggleRefinement to toggleFacetRefinement
it only works on facets.

Haroenv:
original toggleFacetRefinement was renamed to _toggleFacetRefinement (which had 9 occurrences in the code/documentation base). 

Obviously this is quite "optimistic", since it changes the names of an API, which I guess is a breaking change and this not really preferred and maybe that should be avoided. Not 100% sure about it.

Or all occurrences of _toggleFacetRefinement in the docs should be replaced by toggleFacetRefinement, since it should do the same thing except add some extra checks.

fixes #447


bobylito:
First I would to thank you a lot for this PR :) This is the start of a great collaboration :D Will review ASAP :)

Haroenv:
I couldn't wait 😜

dist/algoliasearch.helper.js
```diff
@@ -10623,8 +10623,8 @@ var lib = {
    * @param {string} value the value of the refinement
    * @return {RefinementList} a new and updated list
    */
-  toggleRefinement: function toggleRefinement(refinementList, attribute, value) {
```
bobylito: What about we still implement `toggleRefinement` as a deprecated method? This way, current implementation won't break. If people were using `toggleFacetRefinement` then it's ok because it will automatically trigger `_toggleFacetRefinement` for conjunctive facets.

dist/algoliasearch.helper.js
```diff
@@ -11929,13 +11929,13 @@ SearchParameters.prototype = {
    * @param {value} value value used for filtering
    * @return {SearchParameters}
    */
-  toggleFacetRefinement: function toggleFacetRefinement(facet, value) {
+  _toggleFacetRefinement: function _toggleFacetRefinement(facet, value) {
```
bobylito: Not a big fan of having this function renamed with an underscore and still documenting it. What about renaming it to `toggleConjunctiveFacetRefinement`?

Haroenv: That's a good idea, I didn't think of that as a possibility

dist/algoliasearch.helper.js
bobylito: An example of deprecated method. https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/algoliasearch.helper.js#L307-L312 We keep them in order not to break the current implements and we'll drop them in the next version.

Haroenv: Which version will deprecate it? I picked 2.19.0 now since that's the next major.

src/SearchParameters/RefinementList.js
```diff
@@ -74,8 +74,8 @@ var lib = {
    * @param {string} value the value of the refinement
    * @return {RefinementList} a new and updated list
```
Haroenv: Should this be deprecated too?

dist/algoliasearch.helper.js
bobylito: That should be fine indeed. :)

dist/algoliasearch.helper.js
```diff
@@ -11909,12 +11909,13 @@ SearchParameters.prototype = {
    * @param  {string} value the associated value
    * @return {SearchParameters}
    * @throws will throw an error if the facet is not declared in the settings of the helper
+   * @deprecated since version 2.19.0, see {@link AlgoliaSearchHelper#toggleConjunctiveFacetRefinement}
```
bobylito: This should not be deprecated. :)

Haroenv: Ah, I misunderstood which one that was deprecated. Fixed now

src/SearchParameters/RefinementList.js
bobylito: No. This is the abstract set of functions that handle the structures that manages the facets. Actually this should not even be renamed (see removeRefinement). In this context, it is "clear" that we are not dealing with anything else than facets.

Haroenv: hah, my "replace all" was a bit too powerful maybe. 

So which function should be deprecated then? I think this is the one you commented on in the built version

bobylito: In the SearchParameters:

 - toggleRefinement: should stay, be deprecated, and should be an alias to the new toggleFacetRefinement which will have the same implementation
 - toggleFacetRefinement (the old method): should be renamed to toggleConjunctiveFacetRefinement

In the SearchParameters:

 - toggleRefinement: should be renamed toggleFacetRefinement, and we should still provide a deprecated alias from toggleRefinement to toggleFacetRefinement

There are no changes needed in RefinementList.

Let me know if you have questions on all that :)

Haroenv:
So I switched the deprecation around to a new "toggleFacetRefinement". Could be that I forget the ones in documentation.

bobylito:
Thanks @Haroenv :) Having a look.

dist/algoliasearch.helper.js
```diff
@@ -11908,13 +11908,24 @@ SearchParameters.prototype = {
    * @param  {string} facet the facet to refine
    * @param  {string} value the associated value
    * @return {SearchParameters}
-   * @throws will throw an error if the facet is not declared in the settings of the helper
```
bobylito: It still throws :)

src/algoliasearch.helper.js
```diff
@@ -587,18 +587,18 @@ AlgoliaSearchHelper.prototype.toggleExclude = function() {
  * @fires change
  * @chainable
  */
-AlgoliaSearchHelper.prototype.toggleRefinement = function(facet, value) {
```
bobylito: Can you add an alias method from toggleRefinement to toggleFacetRefinement? So that implementations using toggleRefinement don't break.

dist/algoliasearch.helper.js
Haroenv: fixed locally, will push when not on a train

src/algoliasearch.helper.js
Haroenv: Sure, done

src/SearchParameters/index.js
```diff
@@ -1149,13 +1149,24 @@ SearchParameters.prototype = {
    * @param  {string} facet the facet to refine
    * @param  {string} value the associated value
    * @return {SearchParameters}
-   * @throws will throw an error if the facet is not declared in the settings of the helper
+   * @deprecated since version 2.19.0, see {@link AlgoliaSearchHelper#toggleConjunctiveFacetRefinement}
```
Haroenv: I think this link is wrong, should this be SearchParameters#toggleConjunctive... ?

bobylito: Yes indeed!

dist/algoliasearch.helper.js
bobylito: 👌

bobylito:
Thanks @Haroenv :)

Haroenv:
With loads of pleasure. I want to be sure that the `toggleRefinement` that still exists in the other file is documented correctly, I'll have a check

Haroenv:
@bobylito, there was no conflict with `toggleRefinement` in `RefinementList`, because it isn't explicitly documented, so there's no problem made.

PR #464, opened 2017-01-12
Update/logo

LukyVj:
@bobylito updated :)


documentation-src/metalsmith/layouts/common/footer.jade
```diff
@@ -8,7 +8,7 @@ section.ac-footer
       a.ac-footer-link-item(href='https://github.com/algolia/algoliasearch-helper-js/issues') issues
   .ac-footer-container.ac-footer-brand
     p This project is part of
-    img.ac-footer-brand-icon(src='https://res.cloudinary.com/hilnmyskv/image/upload/logo-community.svg')
+    img.ac-footer-brand-icon(src='images/community-badge.svg')
```
bobylito: why don't we host that on the CDN?

LukyVj: Because kevin said not to, unless it's in the header

PR #474, opened 2017-02-14
fix(agent): sets the helper agent once

### LINKED ISSUE #473: Triple check the client version to prevent adding multiple the UA multiple times

bobylito:
fix #473


bobylito:
Reviewed by @redox :)

PR #479, opened 2017-02-27
feat(search-response): Exposed raw response

proudlygeek:
This PR aims to expose the content `mainSubResponse` inside `SearchResults` in order to have the raw Algolia JSON response available in our JS application.

Here's some more context about it: in our dashboard we would like to display a raw Algolia JSON response like the following:

image

Currently this is not easily achievable by solely using the Helper, as the original, pre-parsed request is not available.

I'm not sure this is the best way to go or if it would be better to optionally expose the field. Any thoughts on this @bobylito?

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -224,6 +224,8 @@ function findMatchingHierarchicalFacetFromAttributeName(hierarchicalFacets, hier
 function SearchResults(state, results) {
   var mainSubResponse = results[0];
 
+  this._rawResults = results;
+
   /**
    * query used to generate the results
    * @member {string}
--- a/test/spec/search.testdata.js
+++ b/test/spec/search.testdata.js
@@ -199,6 +199,7 @@ function getData() {
   });
 
   var responseHelper = {
+    '_rawResults': response.results.slice(0),
     '_state': searchParams,
     'query': '',
     'hits': [


src/SearchResults/index.js
```diff
@@ -224,6 +224,8 @@ function findMatchingHierarchicalFacetFromAttributeName(hierarchicalFacets, hier
 function SearchResults(state, results) {
   var mainSubResponse = results[0];
 
+  this._raw = mainSubResponse;
```
bobylito: Since it's gonna be available, what about we add all the results instead of just the first one?
```js
this._rawResults = results;
```

rayrutjes:
I'm a bit worried about the impact on memory.
Wouldn't it be better to be able to somehow "intercept" the raw response instead of storing in a member variable?

This is more of a debugging / advanced usage feature. It is nice to be able to "inspect" the raw response, but storing it feels wrong.

bobylito:
thanks @rayrutjes for raising the concern. The memory will be allocated for the response anyway, it might just delay the GC (as long as you don't hold a reference on the SearchResults or the ._rawResults)

rayrutjes:
I didn't realize we were only storing a reference to the object `results`. All good ;)

bobylito:
Thanks a lot @proudlygeek I'll keep you updated for the release!

@rayrutjes keep this kind of discussion coming they are important for the sanity of our code!

PR #481, opened 2017-03-07
doc(places): Remove/replace references to Algolia Places

pixelastic:
Mostly this is changing the `utm` in the footer to a specific one for
the Helper, not Algolia places.

Also removed some legacy CSS from the Places website.


Haroenv:
can you run `npm run doc:staging`?

pixelastic:
Actually no. `surge not found`.
Should I add it to `package.json` and create a surge account?

bobylito:
Yes you need to create a surge account for this. However let me have a look.

bobylito:
I will stage it!

bobylito:
http://extra-small-insurance.surge.sh/

PR #482, opened 2017-03-08
Provide an ES module build

rayrutjes:
This makes it easier to include the helper par of another project using es modules.

By adding the `module` and `jsnext:main` key to the package.json, we also simplify integration with rollup and webpack.


Haroenv:
eslint should ignore the `dist` folder

scripts/rollup.esm.config.js
```diff
@@ -0,0 +1,11 @@
+const commonjs = require('rollup-plugin-commonjs');
+
+export default {
+  entry: 'dist/algoliasearch.helper.js',
```
Haroenv: shouldn't this be `index.js`

rayrutjes:
Perfect, 30kb smaller uncompressed and 2kb gzipped 🎉

bobylito:
That's wonderful! thanks :)

vvo:
For the sake of clarity (because I wanted to understand it fully myself): this PR makes using the helper with ES6 only bundlers (= rollup is most probably the target) easier by providing a module that exposes using ES export semantics. This can also be solved by the user via http://rollupjs.org/#using-rollup-with-npm. Still, it's good that we do that for our users so we ease the integration and reduce friction.

The helper should already be working out of the box with webpack without any specific configuration (if not I am missing something and we never got feedback). The fact that we now provide an ES6 build that will be read by webpack is good.

The benefit of providing such build can be even more interesting in project where there are different independent modules to import. https://github.com/algolia/instantsearch.js/issues/1725#issuecomment-283030234

In the helper, given the design of the module, you always need to import everything (import helper from 'module').

Haroenv:
exactly @vvo, this is more of a first step into bringing all libraries to es2015 modules

1. have a build target that exposes modules (<-- this)
2. use es2015 modules in helper, client and libs
3. profit

This first step already makes it easier for rollup, since by default it only supports es2015 modules. Webpack supports both formats for now, and the module version is slightly smaller and gives more flexibility to the end user to build their bundle correctly.

PR #483, opened 2017-03-08
Fea/new header

JonasBa:
New community header added to the build and removed the old unnecessary css and html


bobylito:
Thanks @JonasBa that's awesome ! 

For the sake of readability can you not push the doc built? :)

JonasBa:
Yes, ofc @bobylito :) Let me remove that :)

JonasBa:
@bobylito should be good now ;)

bobylito:
http://noisy-page.surge.sh/

documentation-src/build.gulp.js
```diff
@@ -36,58 +38,83 @@ var documentationRoot = path.join(projectRoot, 'documentation');
...
+
+var header = a.communityHeader({
+  menu:{
+    project: {
+      label: "algoliasearch-helper",
+      url: "/algoliasearch-helper-js/"
+    }
+  },
+  sideMenu: [
+    { name: "Getting started", dropdownItems: null, url: "/algoliasearch-helper-js/gettingstarted.html" },
```
bobylito: You should use relative path to the files. It will help local deployment (and not break on surge / staging)

JonasBa: Let me fix that.

JonasBa:
For now we don't force the use of font inside the component, but we could provide base CSS that will provide the right font settings and that way we have it easily synced all over the place \o/

P.S: The whitespaces are different because you have a set `:root` font size to 14px.

bobylito:
@JonasBa Can you fix the CSS of the header and add it to the PR?

documentation-src/build.gulp.js
Haroenv: is there some way to have the local/dev server be `localhost:PORT/algoliasearch-helper` instead, that should be leading to a more similar experience in both cases

Shipow:
+1 :)

bobylito:
@JonasBa Can you work on the requested changes?

bobylito:
The changesets of files is wrong.

PR #484, opened 2017-03-09
feat(pending-search): let the dev know the pending state

### LINKED ISSUE #469: Provide a way to determine if there is a pending search
Determining if there is a pending search can be useful in the case of bad network and we want to let the user know that his/her request is being searched. Right now this requires to manually keep track of the search and result event and create a counter. This might be wrong if any search is dropped because of the order.

This could be done internally, and we could provide two new entries in the API:
 - `isSearching()` on the helper, that returns `true` if the helper has any pending search (taking care of the discarded searches as well)
 - `noMoreSearchPending` event to indicate that all pending searches have been completed.

Open questions:
 - what about other search via `searchOnce` or `searchForFacetValues`?

bobylito:
With this feature, the developper will be able to know what is the state
of the current searches. If there is a pending search, the dev can
decide to display a visual feedback so that the user something is
going on. This is intended to be able to let the user know that there is
an on-going search, and prevent their frustration.

It implements two things:
 - a new method called `isSearchPending` that returns `true` if there is
a search pending.
 - a new event `noMoreSearch` that is fired when all the requests have
been served or cancelled

http://cute-patch.surge.sh/

Fixes #469


src/algoliasearch.helper.js
```diff
@@ -163,22 +176,34 @@ AlgoliaSearchHelper.prototype.getQuery = function() {
...
+  var self = this;
+
+  this._currentNbQueries++;
+
   if (cb) {
     return this.client.search(
       queries,
       function(err, content) {
+        self._currentNbQueries--;
+        if (self._currentNbQueries === 0) self.emit('noMoreSearch');
```
Haroenv: maybe more something like `searchDone`? I think `noMoreSearch` sounds a bit confusing.

bobylito: Let's go for `searchQueueEmpty` as mentionned during lunch :) Ok for you, @Haroenv ?

Haroenv: sounds the best option so far

bobylito:
@clement-leprovost thanks for the feedback, I will use the same method name.

bobylito:
All set :) Thanks for the review :)

Issue #502, opened 2017-05-26
searchQueueEmpty should be emitted after error if an error occured


rayrutjes:
Currently, if an error occurred, the searchQueueEmpty event is triggered before the error event.

In my use case, I try to wrap the helper in a promise, and I listen to the searchQueueEmpty event to determine if I should resolve the promise. I also listen to the error event to reject it.

Given searchQueueEmpty is always triggered before the error event, I'm unable to catch the error.

bobylito:
What is the benefit to transform the events into promises? The semantic is very different and I'm not sure it fits the usage of the JS Helper. Also why do you need to check if there is a pending search?

vvo:
> In my use case, I try to wrap the helper in a promise

What would help is you to explain the context and usecase before we speak about the failed implementation of wrapping into a Promise.

Maybe there's a feature to be added too.

rayrutjes:
See #504

bobylito:
Released in 2.21.1

PR #503, opened 2017-05-27
fix(pending-search): dispatch error event before searchQueueEmpty

rayrutjes:
This is a fix for #502 .
Let me know if you need more context.

Not sure how to test it. Any help would be more than welcome.

Here is what I want to achieve:

```js
new Promise(resolve => {
  if (this._helper.hasPendingRequests() === false) {
    return resolve();
  }

  // Todo: we need to de-register the one that is not being triggered.
  this._helper.once("searchQueueEmpty", () => {
    resolve();
  });
  this._helper.once("error", error => {
    throw new Error(error.message);
    // Todo: implement rejection once this has a solution
    // https://github.com/algolia/algoliasearch-helper-js/issues/502
    // reject(error);
  });
});
```

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -1227,23 +1227,20 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function(states, queryI
 
   if (err) {
     this.emit('error', err);
-    if (this._currentNbQueries === 0) this.emit('searchQueueEmpty');
-    return;
+  } else {
+    var results = content.results;
+    forEach(states, function(s) {
+      var state = s.state;
+      var queriesCount = s.queriesCount;
+      var helper = s.helper;
+      var specificResults = results.splice(0, queriesCount);
+
+      var formattedResponse = helper.lastResults = new SearchResults(state, specificResults);
+      helper.emit('result', formattedResponse, state);
+    });
   }
-
+  
   if (this._currentNbQueries === 0) this.emit('searchQueueEmpty');
-
-  var results = content.results;
-  forEach(states, function(s) {
-    var state = s.state;
-    var queriesCount = s.queriesCount;
-    var helper = s.helper;
-
-    var specificResults = results.splice(0, queriesCount);
-
-    var formattedResponse = helper.lastResults = new SearchResults(state, specificResults);
-    helper.emit('result', formattedResponse, state);
-  });
 };
 
 AlgoliaSearchHelper.prototype.containsRefinement = function(query, facetFilters, numericFilters, tagFilters) {
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -1223,14 +1223,15 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function(states, queryI
   }
 
   this._currentNbQueries -= (queryId - this._lastQueryIdReceived);
-  if (this._currentNbQueries === 0) this.emit('searchQueueEmpty');
-
   this._lastQueryIdReceived = queryId;
 
   if (err) {
     this.emit('error', err);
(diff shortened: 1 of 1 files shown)


bobylito:
This is stalled until the bug is explicited. Imho there is no bug.

vvo:
In your example you could do:

```
new Promise(resolve => {
  let hadError = false;
  if (this._helper.hasPendingRequests() === false) {
    return resolve();
  }

  // Todo: we need to de-register the one that is not being triggered.
  this._helper.once("searchQueueEmpty", () => {
    Promise.resolve().then(() => {
      if (hadError === false) {
        resolve();
      }
    })
  });
  this._helper.once("error", error => {
    hadError = true;
    reject(error);
  });
});
```

Hacky but should work, Promise.resolve() is similar to setTimeout(fn, 0). If former doesnt work, try the later.

rayrutjes:
Hey, just to be more explicit, I'm not sure this could qualify as a bug.
In my mind it was just an enhancement.

For the explanation in addition to my code above, what I'm trying to do is, at a given point in time, wait for the pending queries to be resolved. But if an error occurs, I want to do something else.

@vvo appreciate the effort there. But it indeed looks hackish ;p. If we could solve it in the helper it would be so much cleaner.

@bobylito could you maybe explain why the PR is an issue? I might be missing some logic.
Is there maybe an alternate option for achieving my goals?

bobylito:
> This is a fix for #502 .

> Hey, just to be more explicit, I'm not sure this could qualify as a bug.

That's why I thought it was a bug and you trying to fix this with this PR. Looking at the code I don't get what you're trying to fix or enhance. As @vvo said, it would help to know what you want to achieve, beyond what you want to code.

As a side note, I'm not sure about translating events into promises. I think that generators are a more fit abstraction for events. (only IMHO and as long as I don't get what is the underlying purpose :))

rayrutjes:
I'm trying to find a way to  `wait until there are no more pending requests`.

My initial idea was just to listen for the `searchQueueEmpty` event but I also need to detect if any error happened.

To say it differently, the moment the queue becomes empty and all calls to the API have been done, I need to know if the state of the helper is reliable or if errors happened.

This PR addresses my use case because it would trigger an `error` event before the `searchQueueEmpty`, and thus allowing me to know an error occurred and that the helper state is not reliable.

bobylito:
> I'm trying to find a way to wait until there are no more pending requests.

Why? :)

> This PR addresses my use case because it would trigger an error event before the searchQueueEmpty, and thus allowing me to know an error occurred and that the helper state is not reliable.

This is interesting. Let's say we define "not reliable" as the moment in which the state triggered an error. The error can either be because of:
 - a network error
 - a configuration issue

In the case of a network error, then it is not that the state is not reliable, it is out of sync with the displayed parameters. This out of sync aspect happens every time a request is sent, no matter what.

For the configuration issue, this can only happen if the user uses `setQueryParameter` basically using any query time configuration or using raw parameters. I think that in both cases that would be solve at dev time.

src/algoliasearch.helper.js
```diff
@@ -1223,15 +1223,16 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function(states, queryI
...
-
   this._lastQueryIdReceived = queryId;
 
   if (err) {
     this.emit('error', err);
+    if (this._currentNbQueries === 0) this.emit('searchQueueEmpty');
     return;
   }
 
+  if (this._currentNbQueries === 0) this.emit('searchQueueEmpty');
```
rayrutjes: Yes, actually this line is at a lot of places in the code. It could be refactored.
In this function, I went the easy path, and didn't want to change too much to ease the discussion by limiting the diff.

If you have a better way to approach this, please feel free to adapt.

If you think my approach could lead to problems, let's chat about it sometime.

There is no big hurry.

rayrutjes:
Closing in favor of #504

PR #504, opened 2017-05-29
misc(event-order): single event statement

bobylito:
Move code around to make sure that we both have a proper event order with `searchQueueEmpty` after `result` or `error`.

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -1227,23 +1227,20 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function(states, queryI
 
   if (err) {
     this.emit('error', err);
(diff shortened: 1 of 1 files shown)


vvo:
👌  should we add a test?

rayrutjes:
LGTM 🎉

rayrutjes:
Do you think we could move forward with this @bobylito ?

bobylito:
For the prosperity, I'm writing why we're moving forward with this:

This changes the order of the event, now `searchQueueEmpty` will be fired after `result` or `error`. The original use case of `searchQueueEmpty` is to provide a way to create a loading indicator, and this change doesn't impact this usage. Also this is a pretty new event so it's quite unlikely that we're gonna have someone already using this event and expecting the previous behaviour.

For those who are, please forgive us ❤️

PR #505, opened 2017-06-01
docs(jsdoc): fix spelling

Haroenv:
I have spelling detection enabled in my editor, and it complained that there were a few typos, so I fixed them.

There's one typo I can't really fix or I'm not sure if it's worth it since it changes a parameter.

There's `...Sizefor...` in a few attributes that should be `...SizeFor...`.

### DIFF
--- a/documentation-src/metalsmith/content/gettingstarted.md
+++ b/documentation-src/metalsmith/content/gettingstarted.md
@@ -278,8 +278,8 @@ values that are meaningful for the current other parameters. Try typing
 This kind of facets are called *conjunctive facets* but they are not the only
 kind of filtering that you can apply with the helper. You can also do:
 
- - disjunctive facetting (for making multiple choices filters)
- - hierarchical facetting (for making hierarchical navigations)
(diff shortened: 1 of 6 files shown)


Haroenv:
Just ran the test locally, and it passes (wow just 4 seconds for such a huge codebase, props!) Since I didn’t change code, it’s probably okay anyway 😄

bobylito:
You're right let's merge. From what I've seen in the debug I would say there is a problem with phantom js and node 8 :/

Haroenv:
@bobylito, I think @vvo pinned the version to node 7 in instantsearch.js, maybe the same?

bobylito:
Maybe :)

vvo:
Yes, easiest is to:

remove version or stable from travis.yml,
put specific version in package.json and nvmrc.

See https://github.com/algolia/react-instantsearch/commit/afc1b2f7e21bbddedc64df70cf5578f3bb95cbf6

Node V8 is very new and some packages might not be installable for now.

PR #514, opened 2017-07-04
fix(events): events for all kinds of searches

### LINKED ISSUE #513: Events for all kind of new search
In 2.20 we added searchQueueEmpty which is great to figure out if there is any search pending. This takes into account all the searches. However we can't get the info about searches for `searchForFacetValues`, `searchOnce`.

We should add events when a search is triggered by those methods with the name of the method as the name of the event:
 - `searchForFacetValues`
 - `searchOnce`

In a later version, we could introduce scoped events but that would be a breaking change as search is already used.

Even though it sounds like a feature, it's actually something that prevents us from implementing a correct loading indicator, which is the original purpose of the new event >> bug.

bobylito:
fix #513

### DIFF
--- a/documentation-src/metalsmith/content/reference.md
+++ b/documentation-src/metalsmith/content/reference.md
@@ -12,7 +12,7 @@ during the search cycle with events:
  - `search`: when the search is sent to Algolia
  - `result`: when the results are retrieved from Algolia
  - `error`: when Algolia sends back an error
- - `searchQueueEmpty`: when there is no more on going searches
+ - `searchQueueEmpty`: when there is no more pending searches
  - `searchForFacetValues`: when a search is sent to Algolia using `searchForFacetValues`
  - `searchOnce`: when a search is sent to Algolia using `searchOnce`
 
--- a/documentation-src/metalsmith/content/reference.md
+++ b/documentation-src/metalsmith/content/reference.md
@@ -13,6 +13,8 @@ during the search cycle with events:
  - `result`: when the results are retrieved from Algolia
  - `error`: when Algolia sends back an error
  - `searchQueueEmpty`: when there is no more on going searches
+ - `searchForFacetValues`: when a search is sent to Algolia using `searchForFacetValues`
+ - `searchOnce`: when a search is sent to Algolia `searchOnce`
 
 You can also read the current parameters of the search using the AlgoliaSearchHelper
 but it might not be the one you expect according to the last results received.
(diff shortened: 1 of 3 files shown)


documentation-src/metalsmith/content/reference.md
```diff
@@ -13,6 +13,8 @@ during the search cycle with events:
  - `result`: when the results are retrieved from Algolia
  - `error`: when Algolia sends back an error
  - `searchQueueEmpty`: when there is no more on going searches
+ - `searchOnce`: when a search is sent to Algolia using `searchForFacetValues`
```
Haroenv: this is the other way around with L17

src/algoliasearch.helper.js
```diff
@@ -180,6 +202,8 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
 
   this._currentNbQueries++;
 
+  this.emit('searchOnce');
```
Haroenv: Does this not need to be tested?

bobylito: Good question :) I guess we could test that! :)

documentation-src/metalsmith/content/reference.md
bobylito: arg!

vvo:
Should we also add the corresponding search states like the search event does? Does it makes sense?

bobylito:
> Should we also add the corresponding search states like the search event does? Does it makes sense?

What do you mean?

src/algoliasearch.helper.js
bobylito: Now it is tested 👍

Haroenv: Don't forget to enjoy your holiday @bobylito

documentation-src/metalsmith/content/reference.md
```diff
@@ -13,6 +13,8 @@ during the search cycle with events:
  - `result`: when the results are retrieved from Algolia
  - `error`: when Algolia sends back an error
  - `searchQueueEmpty`: when there is no more on going searches
```
Haroenv: I'm seeing this now, while it was made earlier, but this should be `ongoing` or `pending`

bobylito: For me it's the same, but please which is which? #lostInTranslation

Haroenv: ongoing means that there are searches that are busy, while pending means that there's something to happen with them still. Both almost the same

bobylito:
@vvo I added the data to the events

vvo:
LGTM

PR #515, opened 2017-07-12
Fix example in SearchResults/index.js

ModuloM:
Add missing `}` in example [in SearchResults section](https://github.com/algolia/algoliasearch-helper-js/blob/789f0430bd2945f93bba04a5540355abfdb889fb/src/SearchResults/index.js#L605).


bobylito:
I need to fix the build :/

ModuloM:
Ouch...I didn't see this coming, sorry.

PR #517, opened 2017-07-20
fix(url): When there are no "other attributes" should not render last &

bobylito:
Before the query string generated was: `query=a&`
now: `query=a`


src/url.js
```diff
@@ -20,6 +20,7 @@ var mapValues = require('lodash/mapValues');
 var isString = require('lodash/isString');
 var isPlainObject = require('lodash/isPlainObject');
 var isArray = require('lodash/isArray');
+var isEmpty = require('lodash/isEmpty');
```
Haroenv: lodash/isEmpty is big compared to other ones, or did you import it in other places already?

PR #523, opened 2017-07-24
chore(tooling): conventional changelog integration

bobylito:
This is a first step towards a better integration of the release step. With that we'll have a way to update the changelog automatically :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "bulkify": "^1.2.0",
     "chalk": "^1.1.3",
     "collect-json": "^1.0.8",
+    "conventional-changelog-cli": "^1.3.2",
     "dateformat": "^1.0.12",
     "dmd": "^1.4.1",
     "doctoc": "^1.2.0",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,9 @@
     "test-node": "clear && node test/run",
     "test:watch": "onchange '{src,test}/**/*.js' 'index.js' -- npm run test-node",
     "test:perf": "node ./test/perf/index.js",
-    "release": "git clean dist/ -f && git checkout dist/ && npm install && ./scripts/release.js && npm run build"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -23,7 +23,9 @@
     "test-node": "clear && node test/run",
     "test:watch": "onchange '{src,test}/**/*.js' 'index.js' -- npm run test-node",
     "test:perf": "node ./test/perf/index.js",
-    "release": "git clean dist/ -f && git checkout dist/ && npm install && ./scripts/release.js && npm run build"
+    "release": "git clean dist/ -f && git checkout dist/ && npm install && ./scripts/release.js && npm run build",
```
vvo: where's release.js?

scripts/conventional-changelog/index.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+var parserOpts = {
```
vvo: I never had to pass a script to conventional-changelog, what are the benefits here from the regular conventional-changelog usage?

package.json
bobylito: This was not touched, I just added a `,`

scripts/conventional-changelog/index.js
bobylito: Seamless integration with 2 years of changelog.

vvo: Oh.. So you are reading back the commit history and generating the changelog out of it somehow?

On some projects what I did is that if the CHANGELOG + state of the project is mixed of good commits and bad commits then I'll configure conventional-changelog to read the **new** commits well and only append to the CHANGELOG file (that I'll rename to CHANGELOG.md). And that's it, we can keep this script but I was a bit surprised to see it (seemed redundant with conventional-changelog defaults)

PR #525, opened 2017-07-25
chore(release): new release script

bobylito:


bobylito:
The best way to test it for real, will be in an actual release :)

package.json
```diff
@@ -96,9 +97,11 @@
     "opn-cli": "^3.1.0",
     "phantomjs-prebuilt": "^2.1.7",
     "pretty-bytes-cli": "^2.0.0",
+    "prompt": "^1.0.0",
     "replace-ext": "1.0.0",
     "rimraf": "^2.5.2",
     "semver": "^5.1.0",
+    "shelljs": "^0.7.8",
```
iam4x: What is the benefit of using `shelljs` instead of:

```js
const {execSync} = require('child_process');
execSync('cat package.json | json');
```

scripts/release.js
```diff
@@ -1,35 +1,136 @@
...
+
+  prompt.message = '?'.info;
+  prompt.colors = false;
+  prompt.start();
+  prompt.get([{
+    description: 'Enter the next version based on the changelog',
+    required: true,
+    conform: function(nextVersion) {
+      return semver.valid(nextVersion) && semver.gte(nextVersion, currentVersion);
+    },
```
iam4x: Nice usage of `semver` will implement in is.js release script 👍

package.json
bobylito: Higher level API + cross platform implementation of common commands. Not fully exploited here but I think it's quite nice :)

bobylito:
Let's roll :)

package.json
vvo: 👍  for shelljs

vvo:
Nice JS script!

PR #526, opened 2017-07-27
fix(SearchResults): Expose exhaustiveNbHits and exhaustiveFacetsCount in SearchResults

### LINKED ISSUE #489: Expose exhaustiveNbHits and exhaustiveFacetsCount in SearchResults
For now each SearchResults exposes `timeoutCounts` and `timeoutHits`.
https://community.algolia.com/algoliasearch-helper-js/reference.html#SearchResults#timeoutCounts

 Those are now depreciated, and have been replaced by 
`exhaustiveNbHits` and `exhaustiveFacetsCount`.

Would be great to have those availabe in the helper's SearchResults object.

bobylito:
Fix #489

### DIFF
--- a/documentation-src/metalsmith/content/reference.md
+++ b/documentation-src/metalsmith/content/reference.md
@@ -475,8 +475,8 @@ are implemented on top of Algolia API such as faceting.
 
 {{> jsdoc jsdoc/results/processingTimeMS}}
 {{> jsdoc jsdoc/results/serverUsed}}
-{{> jsdoc jsdoc/results/timeoutCounts}}
-{{> jsdoc jsdoc/results/timeoutHits}}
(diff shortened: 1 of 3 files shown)


bobylito:
> this will only have an impact if that value is true, right?

Not sure to get what you mean. The purpose is to pass those two attributes from the Algolia response into the SearchResults. There is no underlying logic.

Haroenv:
I mean that, if the result doesn't have `exhaustiveNbHits`, the value of `SearchResults. exhaustiveNbHits` will be `undefined`, right?

bobylito:
> I mean that, if the result doesn't have exhaustiveNbHits, the value of SearchResults. exhaustiveNbHits will be undefined, right?

Yes that's correct. The only case, I can think of, is when we ask for less parameters in the response. Do you see other cases?

PR #542, opened 2017-10-08
fix(setState): use .make() instead of constructor()

iam4x:
**Bug:**

When you call `helper.setState()` it won't compute `hierarchicalFacets` to check if there is a `rootPath` defined to set a `hierarchicalFacetRefinement`.

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -806,7 +806,7 @@ AlgoliaSearchHelper.prototype.setQueryParameter = function(parameter, value) {
  * @chainable
  */
 AlgoliaSearchHelper.prototype.setState = function(newState) {
-  this.state = new SearchParameters(newState);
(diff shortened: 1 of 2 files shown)


bobylito:
You just reached a new level @iam4x 💯 

level up

iam4x:
Thanks @bobylito I've fixed the tests it should be good 👍

bobylito:
> Thanks @bobylito I've fixed the tests it should be good 👍

What did you have to change to the tests?

iam4x:
missing `t.end()` see the last commit (https://github.com/algolia/algoliasearch-helper-js/pull/542/commits/027ce5672ae6af6783c4375b31d4c9d004902b01)

bobylito:
> missing t.end() see the last commit (027ce56)

Classic!

PR #543, opened 2017-10-09
fix(requestBuilder): set analytics:false to subsequent queries

### LINKED ISSUE #540: subsequent requests for disjunctive facets should use analytics=false
When we’re performing a disjunctive query, we’re sending two queries: one to generate the results, and a secondary one to generate the facet values (and potentially more than two queries if there are multiple disjunctive facets).

One issue that we have is that the secondary queries are present in the analytics, even though they shouldn’t.

The fix for this one is pretty easy: all of the secondary queries generated by the disjunctive facets should have the search parameter analytics=false (https://www.algolia.com/doc/api-reference/api-parameters/analytics/).

https://app.asana.com/0/214000806178027/442490537062541 (internal)

bobylito:
Fix #540

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -96,7 +96,8 @@ var requestBuilder = {
       attributesToRetrieve: [],
       attributesToHighlight: [],
       attributesToSnippet: [],
-      tagFilters: tagFilters
+      tagFilters: tagFilters,
+      analytics: false
     };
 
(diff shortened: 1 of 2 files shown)


bobylito:
> there's a FIXME about the rootPath? What can we do about this one?

Well if you have a better understanding of what this variable means, please help 🙇

PR #547, opened 2017-10-10
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that algoliasearch-helper-js gets 24M hits per month on jsDelivr, so I thought you might like this badge.


bobylito:
Thanks for adding this @LukasDrgon

PR #548, opened 2017-10-11
chore(doc): export the svg for the principles without text

bobylito:


bobylito:
I just updated the images here it should be fine :)

PR #549, opened 2017-10-18
feat(sffv): can override search when using searchForFacetValues

bobylito:
Until now it was not possible to update the state before searching for a facet value. But this can be annoying to not be able the search state that is sent to Algolia. This PR adds a new parameter to the method.

It does make for a great looking API but I didn't want to make a breaking change.

This PR is required for https://github.com/algolia/instantsearch.js/issues/2474

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -262,14 +262,16 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
  * See the description of [FacetSearchResult](reference.html#FacetSearchResult)
  * @param {string} facet the name of the faceted attribute
  * @param {string} query the string query for the search
- * @param {number} maxFacetHits the maximum number values returned. Should be > 0 and <= 100
- * @return {promise<FacetSearchResult>} the results of the search
+ * @param {number} [maxFacetHits] the maximum number values returned. Should be > 0 and <= 100
+ * @param {object} [userState] the set of custom parameters to use on top of the current state. Setting a property to `undefined` removes
+ * it in the generated query.
+ * @return {promise.<FacetSearchResult>} the results of the search
  */
-AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxFacetHits) {
-  var state = this.state;
-  var index = this.client.initIndex(this.state.index);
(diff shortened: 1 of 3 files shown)


src/algoliasearch.helper.js
```diff
@@ -262,14 +262,16 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
...
  * @param {string} query the string query for the search
- * @param {number} maxFacetHits the maximum number values returned. Should be > 0 and <= 100
- * @return {promise<FacetSearchResult>} the results of the search
- */
-AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxFacetHits) {
-  var state = this.state;
-  var index = this.client.initIndex(this.state.index);
+ * @param {number} [maxFacetHits] the maximum number values returned. Should be > 0 and <= 100
+ * @param {object} [userState] the set of custom parameters to use on top of the current state. Setting a property to `undefined` removes
+ * it in the generated query.
```
iam4x: We should maybe accept `null` value as well?

iam4x: Oh reading next line:

```js
var state = this.state.setQueryParameters(userState || {});
```

We can maybe update the documentation to:

> Setting a falsy property (`null`, `false`, `undefined`, `''`) removes it in the generated query

test/spec/algoliasearch.helper/searchForFacetValues.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+var test = require('tape');
+// var algoliaSearch = require('algoliasearch');
+// var SearchParameters = require('../../../src/SearchParameters');
```
iam4x: Left-over?

bobylito: 👌

src/algoliasearch.helper.js
Haroenv: I don't think we should care about the variance here. Let's keep it `undefined` IMO

bobylito: Ok so interestingly, `undefined` is the only falsy value that let you unset a state property.

PR #550, opened 2017-10-20
Replace some lodash methods by native methods [WIP]

Haroenv:
This should be tested thoroughly!

35.1 KB --> 34.7 KB

### DIFF
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -18,7 +18,6 @@ var isFunction = require('lodash/isFunction');
 var isEmpty = require('lodash/isEmpty');
 var defaults = require('lodash/defaults');
 
-var reduce = require('lodash/reduce');
 var filter = require('lodash/filter');
 var omit = require('lodash/omit');
 
@@ -104,7 +103,7 @@ var lib = {
     } else if (isFunction(attribute)) {
       var hasChanged = false;
 
-      var newRefinementList = reduce(refinementList, function(memo, values, key) {
+      var newRefinementList = refinementList.reduce(function(memo, values, key) {
         var facetList = filter(values, function(value) {
           return !attribute(value, key, refinementType);
         });
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -131,8 +131,6 @@ var lib = {
    * @return {boolean}
    */
   isRefined: function isRefined(refinementList, attribute, refinementValue) {
-    var indexOf = require('lodash/indexOf');
-
     var containsRefinements = !!refinementList[attribute] &&
       refinementList[attribute].length > 0;
 
@@ -142,7 +140,7 @@ var lib = {
 
     var refinementValueAsString = '' + refinementValue;
 
-    return indexOf(refinementList[attribute], refinementValueAsString) !== -1;
+    return refinementList[attribute].indexOf(refinementValueAsString) !== -1;
   }
 };
 
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -103,13 +103,13 @@ var lib = {
     } else if (isFunction(attribute)) {
       var hasChanged = false;
 
-      var newRefinementList = refinementList.reduce(function(memo, values, key) {
-        var facetList = filter(values, function(value) {
+      var newRefinementList = Object.keys(refinementList).reduce(function(memo, key) {
+        var facetList = filter(refinementList[key], function(value) {
           return !attribute(value, key, refinementType);
         });
 
         if (!isEmpty(facetList)) {
-          if (facetList.length !== values.length) hasChanged = true;
+          if (facetList.length !== refinementList[key].length) hasChanged = true;
           memo[key] = facetList;
         }
         else hasChanged = true;
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -6,7 +6,6 @@ var forOwn = require('lodash/forOwn');
 var forEach = require('lodash/forEach');
 var filter = require('lodash/filter');
 var omit = require('lodash/omit');
-var indexOf = require('lodash/indexOf');
 var isNaN = require('lodash/isNaN');
 var isEmpty = require('lodash/isEmpty');
 var isEqual = require('lodash/isEqual');
@@ -1339,7 +1338,7 @@ SearchParameters.prototype = {
    * @return {boolean}
    */
   isDisjunctiveFacet: function(facet) {
-    return indexOf(this.disjunctiveFacets, facet) > -1;
+    return this.disjunctiveFacets.indexOf(facet) > -1;
   },
   /**
    * Test if the facet name is from one of the hierarchical facets
@@ -1357,7 +1356,7 @@ SearchParameters.prototype = {
    * @return {boolean}
    */
   isConjunctiveFacet: function(facet) {
-    return indexOf(this.facets, facet) > -1;
+    return this.facets.indexOf(facet) > -1;
   },
   /**
    * Returns true if the facet is refined, either for a specific value or in
@@ -1427,7 +1426,7 @@ SearchParameters.prototype = {
       return refinements.length > 0;
     }
 
-    return indexOf(refinements, value) !== -1;
+    return refinements.indexOf(value) !== -1;
   },
   /**
    * Test if the triple (attribute, operator, value) is already refined.
@@ -1465,7 +1464,7 @@ SearchParameters.prototype = {
    * @return {boolean}
    */
   isTagRefined: function isTagRefined(tag) {
-    return indexOf(this.tagRefinements, tag) !== -1;
+    return this.tagRefinements.indexOf(tag) !== -1;
   },
   /**
    * Returns the list of all disjunctive facets refined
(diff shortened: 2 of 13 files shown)


Haroenv:
Are there only 5 tests?

Haroenv:
arg, I forgot that _.map and _.reduce also work on objects; should check if that's the case here...

src/SearchParameters/RefinementList.js
```diff
@@ -104,7 +103,7 @@ var lib = {
     } else if (isFunction(attribute)) {
       var hasChanged = false;
 
-      var newRefinementList = reduce(refinementList, function(memo, values, key) {
+      var newRefinementList = refinementList.reduce(function(memo, values, key) {
```
Haroenv: This might be an Object 🤔

src/SearchParameters/index.js
```diff
@@ -894,7 +890,7 @@ SearchParameters.prototype = {
       return omit(this.numericRefinements, attribute);
     } else if (isFunction(attribute)) {
       var hasChanged = false;
-      var newNumericRefinements = reduce(this.numericRefinements, function(memo, operators, key) {
+      var newNumericRefinements = this.numericRefinements.reduce(function(memo, operators, key) {
```
Haroenv: object maybe?

src/SearchParameters/index.js
```diff
@@ -518,9 +514,9 @@ SearchParameters._parseNumbers = function(partialState) {
     forEach(partialState.numericRefinements, function(operators, attribute) {
       numericRefinements[attribute] = {};
       forEach(operators, function(values, operator) {
-        var parsedValues = map(values, function(v) {
-          if (isArray(v)) {
-            return map(v, function(vPrime) {
+        var parsedValues = values.map(function(v) {
```
Haroenv: object maybe

src/SearchParameters/index.js
```diff
@@ -1499,7 +1495,7 @@ SearchParameters.prototype = {
     return intersection(
       // enforce the order between the two arrays,
       // so that refinement name index === hierarchical facet index
-      map(this.hierarchicalFacets, 'name'),
+      this.hierarchicalFacets.map('name'),
```
Haroenv: object

src/SearchParameters/index.js
```diff
@@ -1693,7 +1689,7 @@ SearchParameters.prototype = {
       this.getHierarchicalFacetByName(facetName)
     );
     var path = refinement.split(separator);
-    return map(path, trim);
+    return path.map(trim);
```
Haroenv: probably an object

src/SearchResults/generate-hierarchical-tree.js
```diff
@@ -31,7 +29,7 @@ function generateTrees(state) {
       results = hierarchicalFacetResult.slice(hierarchicalRootPath.split(hierarchicalSeparator).length);
     }
 
-    return reduce(results, generateTreeFn, {
+    return results.reduce(generateTreeFn, {
```
Haroenv: maybe an object

src/SearchResults/index.js
```diff
@@ -334,7 +332,7 @@ function SearchResults(state, results) {
    * disjunctive facets results
    * @member {SearchResults.HierarchicalFacet[]}
    */
-  this.hierarchicalFacets = map(state.hierarchicalFacets, function initFutureTree() {
+  this.hierarchicalFacets = state.hierarchicalFacets.map(function initFutureTree() {
```
Haroenv: object?

src/SearchResults/index.js
```diff
@@ -510,7 +508,7 @@ function SearchResults(state, results) {
     });
   });
 
-  this.hierarchicalFacets = map(this.hierarchicalFacets, generateHierarchicalTree(state));
+  this.hierarchicalFacets = this.hierarchicalFacets.map(generateHierarchicalTree(state));
```
Haroenv: object?

src/SearchResults/index.js
```diff
@@ -545,7 +543,7 @@ function extractNormalizedFacetValues(results, attribute) {
     var facet = find(results.facets, predicate);
     if (!facet) return [];
 
-    return map(facet.data, function(v, k) {
+    return facet.data.map(function(v, k) {
```
Haroenv: object?

src/SearchResults/index.js
```diff
@@ -557,7 +555,7 @@ function extractNormalizedFacetValues(results, attribute) {
     var disjunctiveFacet = find(results.disjunctiveFacets, predicate);
     if (!disjunctiveFacet) return [];
 
-    return map(disjunctiveFacet.data, function(v, k) {
+    return disjunctiveFacet.data.map(function(v, k) {
```
Haroenv: object?

src/SearchResults/index.js
```diff
@@ -578,7 +576,7 @@ function recSort(sortFn, node) {
   if (!node.data || node.data.length === 0) {
     return node;
   }
-  var children = map(node.data, partial(recSort, sortFn));
+  var children = node.data.map(partial(recSort, sortFn));
```
Haroenv: object?

src/algoliasearch.helper.js
```diff
@@ -1184,7 +1183,7 @@ AlgoliaSearchHelper.prototype._search = function() {
 
   this.emit('search', state, this.lastResults);
 
-  var derivedQueries = map(this.derivedHelpers, function(derivedHelper) {
+  var derivedQueries = this.derivedHelpers.map(function(derivedHelper) {
```
Haroenv: probably an array but maybe object?

src/requestBuilder.js
```diff
@@ -251,8 +248,7 @@ var requestBuilder = {
   _getHitsHierarchicalFacetsAttributes: function(state) {
     var out = [];
 
-    return reduce(
-      state.hierarchicalFacets,
+    return state.hierarchicalFacets.reduce(
```
Haroenv: probably object

src/SearchResults/index.js
vvo: > arg, I forgot that _.map and _.reduce also work on objects; should check if that's the case here...

Object.entries() works well for those cases

Haroenv: Yep, will have to use object.keys in those (entries is edge, end helper supports ie9)

vvo: This should be discussed with everyone about the browser support of helper, IE9 is really old. We can rely on polyfills. Like InstantSearch does

Haroenv: I’d rather not add polyfills in a PR aimed at making the size smaller, but we can go further in removing methods if ie11 or edge is picked as target

src/SearchParameters/RefinementList.js
iam4x: Object here, look the reducer function 👍

src/SearchParameters/index.js
iam4x: Yes object as well

src/SearchParameters/index.js
iam4x: This should be:

```js
this.hierarchicalFacets.map(({name}) => name)
```

src/SearchParameters/index.js
iam4x: Nope, `refinement.split` would return an array 👍

src/SearchResults/generate-hierarchical-tree.js
iam4x: Nope it's an array, worked on that piece 👍

src/SearchResults/index.js
iam4x: Array as well

bobylito:
Hi @Haroenv and thanks a lot for your contribution and reopening the discussion. The move to native methods on collections has been a question for a long time now. This was never because of a few key questions that we had:
 - what is the size benefit that we can expect?
 - what is the impact on the dev experience?
 - what is the impact on performance?
 - can we actually replace lodash completely and at what cost?

Can you go in further details in your analysis on those topics? Maybe you can figure out all the lodash functions used (and based on their usage) and see the one that have zero cost for removal (eg: map for arrays), small (eg: map for objects / collections)  and big (function that don't have a native counterpart).

Also on the perf part, Lodash has always been a library that strives for perf. What is the cost of going native? Are there any browsers where this impact is worse?

The objectives here must be that:
 - the size gain must be substantial
 - we don't want to support a lodash-like
 - the dev experience should not be impacted (it's already complex enough)

bobylito:
Hi @Haroenv and thanks a lot for your contribution and reopening the discussion. The move to native methods on collections has been a question for a long time now. This was never because of a few key questions that we had:
 - what is the size benefit that we can expect?
 - what is the impact on the dev experience?
 - what is the impact on performance?
 - can we actually replace lodash completely and at what cost?

Can you go in further details in your analysis on those topics? Maybe you can figure out all the lodash functions used (and based on their usage) and see the one that have zero cost for removal (eg: map for arrays), small (eg: map for objects / collections)  and big (function that don't have a native counterpart).

Also on the perf part, Lodash has always been a library that strives for perf. What is the cost of going native? Are there any browsers where this impact is worse?

The objectives here must be that:
 - the size gain must be substantial
 - we don't want to support a lodash-like
 - the dev experience should not be impacted (it's already complex enough)

src/SearchResults/index.js
Haroenv: For the record: `this.hierarchicalFacets` is an array

Haroenv:
Let's discuss this in #552 instead. We can easily replicate the work done here later

PR #554, opened 2017-10-24
fix(url): treat insideBoundingBox in float form as number

### LINKED ISSUE #553: insideBoundingBox retrieved as strings from getStateFromQueryString
When you populate your state from a query string and it contains a insideBoundingBox parameter, the search fails with 

> Invalid value for "insideBoundingBox" parameter, expected float

example :
`q=baguette&iBB[0][0]=51.1241999&iBB[0][1]=9.662499900000057&iBB[0][2]=41.3253001&iBB[0][3]=-5.559099999999944`

return 
`{
     ...,
     insideBoundingBox: [
       ["51.1241999", "9.662499900000057", "41.3253001", "-5.559099999999944"]
     ],
   }
`
so using the result directly in a `helper.setState` will fail.

would be great to either accept string representations of floats or transform the query string correctly.

Haroenv:
fixes #553

There's two representation of insideBoundingBox, one is `'47.3165,4.9665,47.3424,5.0201'`, the other is `[[47.3165, 4.9665,47.3424, 5.0201], [40.9234, 2.1185, 38.6430, 1.9916]]`. If URLSync was
used with the second form, it would be parsed as `[ ["51.1241999","9.662499900000057", "41.3253001", "-5.559099999999944"] ]` instead of the original array of array of floats.

This commit adds special handling for this kind of shape to the `_parseNumbers` of `SearchParameters`

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -513,6 +513,16 @@ SearchParameters._parseNumbers = function(partialState) {
     }
   });
 
+  // there's two formats of insideBoundingBox, we need to parse
+  // the one which is an array of float geo rectangles
+  if (Array.isArray(partialState.insideBoundingBox)) {
+    numbers.insideBoundingBox = partialState.insideBoundingBox.map(function(geoRect) {
+      return geoRect.map(function(value) {
+        return parseFloat(value);
+      });
+    });
+  }
+
   if (partialState.numericRefinements) {
     var numericRefinements = {};
     forEach(partialState.numericRefinements, function(operators, attribute) {
--- a/test/spec/SearchParameters/_parseNumbers.js
+++ b/test/spec/SearchParameters/_parseNumbers.js
@@ -8,6 +8,7 @@ test('_parseNumbers should convert to number all specified root keys (that are p
     aroundPrecision: '42',
     aroundRadius: '42',
     getRankingInfo: '42',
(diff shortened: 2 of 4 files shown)


Haroenv:
should work now @bobylito

bobylito:
Inside bounding box is a list of coordinate, so I don't think it should be parsed like a number or am I missing something?

https://www.algolia.com/doc/api-reference/api-parameters/insideBoundingBox/

Haroenv:
Well each coordinate is a number, no?

bobylito:
Well what I see in the doc is that the value is a serie of number separated by commas like `47.3165,4.9665,47.3424,5.0201`.

samouss:
It can be both:

- a serie of number separated by coma: `7.3165,4.9665,47.3424,5.0201`
- an array of array: `[[7.3165, 4.9665, 47.3424, 5.0201]]`

bobylito:
Yes true @samouss so parsing as numbers will not be ok. Since we don't make manage this parameter, I don't we should do anything here. Again I might be missing something :)

bobylito:
@Haroenv is there something I can do to help you make this PR move forward?

Haroenv:
I wasn’t too sure where to start with that test, so maybe you can take a look there @bobylito

bobylito:
> I wasn’t too sure where to start with that test, so maybe you can take a look there @bobylito

Let's meet next week so I can help you getting started 👍

PR #557, opened 2017-11-22
chore(package.json): add files entry

Haroenv:
Right now all files in the directory are published to npm on publish. On the [Yarn site](https://yarn.pm/algoliasearch-helper) you can check which files are included now, which includes the documentation and other files. 

Also removes the .npmignore file, since a whitelist is better than a blacklist

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,0 @@
-examples/
-test/
-# we do not want to lock down dependencies when used as an npm dependency
-# only when we build
-npm-shrinkwrap.json
--- a/package.json
+++ b/package.json
@@ -40,7 +40,6 @@
   "files": [
     "dist",
     "src",
-    "CHANGELOG"
   ],
   "devDependencies": {
     "algolia-frontend-components": "^0.0.33",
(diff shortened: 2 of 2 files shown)


bobylito:
Can you also keep the readme and the changelog. What the bower.json, the package.json and the yarn.lock too?

Haroenv:
readme, changelog and package.json are kept automatically. yarn.lock isn't useful to keep in published packages because it doesn't apply

bobylito:
Sorry @Haroenv I didn't follow up on this one. For the yarn.lock, isn't this file shows the actual version of the dependencies used to build the package? That could be useful information when trying to debug some issues.

Haroenv:
yeah, can be used for debugging, but you can go to the tag on GitHub for that IMO. Also what really counts for someone using the library is not the locked versions, but those in package.json are. Explained in the blog too: https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

Should be merged IMO

flow for finding lock file for a certain version (latest here, all are possible to get though):

bobylito:
Ok good arguments, thanks @Haroenv for your patience 👍

PR #559, opened 2017-12-14
fix(release-script): actually build the library

bobylito:
There is a missing step in the release process: the build step. This patch adds this step using the existing script.


bobylito:
Thanks for the quick review @iam4x

PR #562, opened 2018-01-29
Feat/insights

JonasBa:
Improves helper usage for algolia-insights library by copying queryID to results.
Thanks a lot @bobylito for the guidance here :)!

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -97,7 +97,8 @@ var requestBuilder = {
       attributesToHighlight: [],
       attributesToSnippet: [],
       tagFilters: tagFilters,
-      analytics: false
+      analytics: false,
+      clickAnalytics: false
     };
 
     var hierarchicalFacet = state.getHierarchicalFacetByName(facet);
--- a/test/integration-spec/helper.insights.js
+++ b/test/integration-spec/helper.insights.js
@@ -1,0 +1,54 @@
+'use strict';
+
+var utils = require('../integration-utils.js');
+var setup = utils.setupSimple;
+
+var algoliasearchHelper = utils.isCIBrowser ? window.algoliasearchHelper : require('../../');
+
+var test = require('tape');
+var random = require('lodash/random');
+
+if (!utils.shouldRun) {
+  test = test.skip;
+}
+var indexName = '_travis-algoliasearch-helper-js-' +
+  (process.env.TRAVIS_BUILD_NUMBER || 'DEV') +
+  'helper_searchonce' + random(0, 5000);
+
+var dataset = [
+  {objectID: '1', name: 'TestName'}
+];
+
+var config = {};
+
+test(
+  '[INT][INSIGHTS] search with clickAnalytics should have a queryID',
+  function(t) {
+    setup(indexName, dataset, config).
+    then(function(client) {
+      var helper = algoliasearchHelper(client, indexName, {clickAnalytics: true});
+      helper.on('result', function(content) {
+        t.notEqual(content.queryID, undefined);
+        t.end();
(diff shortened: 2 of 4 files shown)


test/integration-spec/helper.insights.js
```diff
@@ -0,0 +1,54 @@
...
+
+test(
+  '[INT][INSIGHTS] search with clickAnalytics should have a queryID',
+  function(t) {
+    setup(indexName, dataset, config).
+    then(function(client) {
+      var helper = algoliasearchHelper(client, indexName, {clickAnalytics: true});
+      helper.on('result', function(content) {
+        t.notEqual(content.queryID, undefined);
+        t.end();
```
Haroenv: not sure, but could you say it needs to be a string of X length with `tape`?

Haroenv: worst case, `typeof` and `.length`

bobylito: That's a good suggestion for improvement, can you do that @JonasBa ?

src/requestBuilder.js
```diff
@@ -97,7 +97,8 @@ var requestBuilder = {
       attributesToHighlight: [],
       attributesToSnippet: [],
       tagFilters: tagFilters,
-      analytics: false
+      analytics: false,
+      clickAnalytics: false
```
vvo: is it called clickAnalytics or insights?

test/spec/requestBuilder.js
```diff
@@ -16,3 +44,16 @@ test('The request builder should generate queries according to the search parame
...
   t.end();
 });
+
+test('The request builder should should force clickAnalytics to false on subsequent queries if not specified', function(t) {
+  var testData = require('./search.testdata.js')();
+  var searchParams = testData.searchParams;
+
+  var queries = getQueries(searchParams.index, searchParams);
+  t.equal(queries.length, 2);
+  t.equal(queries[0].params.clickAnalytics, undefined, 'the parameter clickAnalytics should be defined on the first query');
```
samouss: Why we leave the value as `undefined` instead of explicitly set to `true` or `false`?

test/integration-spec/helper.insights.js
iam4x: @Haroenv You could also use a regex to test the string and the length 👍 

```js
/\S{2}/.test('ab') // === true
```

test/spec/requestBuilder.js
```diff
@@ -5,7 +5,35 @@ var test = require('tape');
 var requestBuilder = require('../../src/requestBuilder.js');
 var getQueries = requestBuilder._getQueries;
 
-test('The request builder should generate queries according to the search parameters', function(t) {
+test('The request builder should set analytics to subsequent queries', function(t) {
```
iam4x: Just a question, what is the logic of subsequent queries? I'm not sure to understand why the second query should have the param analytics false?

Haroenv: That's to make sure that if a user types one letter, it's not counted as multiple queries, for example in hierarchical faceting

test/spec/requestBuilder.js
bobylito: From the Algolia rest API perspective there is already a default value set to `false`. We force the value to `false` on subsequent because we know for sure that those queries should never be analyzed. On the first query, it's less clear so we leave the engine decides what behavior it should use.

test/integration-spec/helper.insights.js
bobylito: However the length might change and be formatted differently. So maybe we should just test that it's a non empty string.

PR #566, opened 2018-03-09
chore(deps): pin and upgrade some dependencies

bobylito:
Upgraded dependencies: 
 - algoliasearch
 - lodash
 - qs


bobylito:
@Haroenv that's true.

However in this case, we're actually talking about 4 super stable dependencies and since we don't test with all the different version of the libraries we support, it's better to actually make that it works. 

On the other hand, pinning let us enable renovate and automate the update of those libraries.

vvo:
One thing I still dont manage to get my mind around is why the default js:libs config of renovate says "pin only devdependencies" while we are trying to go the other way around in our own libraries.

(cf https://renovateapp.com/docs/config-presets/config-config)

samouss:
https://renovateapp.com/docs/deep-dives/dependency-pinning#why-use-ranges

vvo:
@samouss Can you precise what is your thought on this pinning dependencies or not? Thx

bobylito:
@Haroenv @vvo @Haroenv I've updated the PR to follow what we said. We're good now you think?

package.json
```diff
@@ -43,81 +43,76 @@
...
+    "webpack": "1.15.0",
+    "webpack-stream": "3.2.0",
+    "zuul": "^3.11.1",
+    "zuul-ngrok": "4.0.0"
   },
   "dependencies": {
-    "events": "^1.1.0",
-    "lodash": "^4.13.1",
-    "qs": "^6.2.1",
-    "util": "^0.10.3"
```
samouss: Is it intended to remove this dependencies? Since we are still using it.

https://github.com/algolia/algoliasearch-helper-js/blob/d7985565d2a28e677e7ea56badcefda41a8f5b30/src/DerivedHelper/index.js#L2

package.json
```diff
@@ -43,81 +43,76 @@
...
+    "sinon": "4.4.9",
+    "st": "1.2.0",
+    "tap-spec": "4.1.1",
+    "tape": "4.7.0",
+    "uglify-js": "2.8.29",
+    "vinyl": "1.2.0",
+    "watchify": "3.7.0",
+    "webpack": "1.15.0",
+    "webpack-stream": "3.2.0",
+    "zuul": "^3.11.1",
```
samouss: Sill have a caret.

bobylito:
> with the pinned dependencies we should definitely have renovate here though

That's the plan.

PR #570, opened 2018-03-21
chore(test): delete all generated indices at the end of the tests

### LINKED ISSUE #560: Tests don't delete all indices
It seems like the test suite create indices that are not deleted after the test ends. It results in a lot of indices left in the Algolia app.

Example: 
* _travis-algoliasearch-helper-js-809helper_searchonce1299
* _travis-algoliasearch-helper-js-807helper_sffv807
(numbers change but the rest of the name is always the same.

bobylito:
fix #560

### DIFF
--- a/test/run.js
+++ b/test/run.js
@@ -1,10 +1,26 @@
 'use strict';
 
 var bulk = require('bulk-require');
+var test = require('tape');
+var algolia = require('algoliasearch');
 
 bulk(__dirname, ['spec/**/*.js']);
 
 if (process.env.ONLY_UNIT !== 'true' && process.env.INTEGRATION_TEST_API_KEY && process.env.INTEGRATION_TEST_APPID) {
   // usage: INTEGRATION_TEST_APPID=$APPID INTEGRATION_TEST_API_KEY=$APIKEY npm run dev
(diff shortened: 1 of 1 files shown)


test/run.js
```diff
@@ -1,10 +1,26 @@
...
+}
+
+function cleanupIntegration() {
+  console.log('Deleting all indices');
+  var client = algolia(process.env.INTEGRATION_TEST_APPID, process.env.INTEGRATION_TEST_API_KEY);
+  var maybeIndices = client.listIndexes();
+  maybeIndices.then(content => {
+    content.items
+      .map(i => i.name)
+      .filter(n => n.indexOf('_travis-algoliasearch-helper') !== -1)
```
Haroenv: you can also use `n.includes('_travis-algoliasearch-helper')` ([String#includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes))

Haroenv:
@julienbourdeau, we should be able to use the api key generator right?

test/run.js
bobylito: I'm old school :P

PR #571, opened 2018-03-30
feat(search): Promisify `client.search()`

francoischalifour:
_This PR is part of a series of improvements in our JavaScript Helper and Client._

-----

This feature unifies the helper's internal API by using the promisified version of `client.search()` in the `helper.search()` method.

## Why

As we're going towards a more backend agnostic InstantSearch, we want to unify our APIs. The `searchForFacetValues()` method supports promises while the current `search()` method uses callbacks. We want the `search()` method to be aligned with the SFFV method: only use promises.

This is an important improvement for custom backend implementations: all our search methods would now use promises, making the API easier to understand.

## What

* `helper._search()` uses the promisified version of `client.search()`
* `sinon` has been upgraded to use `sinon.stub().resolves()` (useful for this new feature)
* Tests have been updated with new implementation

## Related

algolia/algoliasearch-client-javascript#677

-----

_Another PR will be sent to use the top-level `client.searchForFacetValues()` once algolia/algoliasearch-client-javascript#677 is merged._


src/algoliasearch.helper.js
```diff
@@ -1201,7 +1201,12 @@ AlgoliaSearchHelper.prototype._search = function() {
 
   this._currentNbQueries++;
 
-  this.client.search(queries, this._dispatchAlgoliaResponse.bind(this, states, queryId));
+  this.client.search(queries)
+    .then(this._dispatchAlgoliaResponse.bind(this, states, queryId))
+    .catch(this._dispatchAlgoliaError.bind(this, queryId))
+    .catch(function(err) {
+      console.error('error', err);
```
francoischalifour: I'm not sure logging the error is the best thing to do at this point.

Haroenv: `AlgoliaSearchHelper.emit('error', err)` seems to be closer to the original version of this code

Haroenv: oh didn't read enough, I think actually the one catch should be enough

test/spec/algoliasearch.helper/pendingSearch.js
```diff
@@ -136,8 +138,10 @@ test('When helper.search() and one request is discarded, hasPendingRequests is t
   var client = algoliaSearch('dsf', 'dsfdf');
 
   var triggerCbs = [];
-  client.search = function(qs, cb) {
-    triggerCbs.push(function() { cb(null, testData().response); });
+  client.search = function() {
```
francoischalifour: This test doesn't work for now.

src/algoliasearch.helper.js
francoischalifour: With only the first `catch`, you can get an `UnhandledPromiseRejectionWarning`.

Haroenv: maybe then just also emit an error?

francoischalifour: You're right! I'll do it at a higher level with a `try` `catch`.

test/spec/algoliasearch.helper/pendingSearch.js
```diff
@@ -5,35 +5,6 @@ var algoliaSearch = require('algoliasearch');
 
 var algoliasearchHelper = require('../../../index');
 
-test('When searchOnce with callback, hasPendingRequests is true', function(t) {
```
francoischalifour: I removed this test because `search()` with callbacks doesn't exist anymore. Let me know if I'm wrong @bobylito.

francoischalifour:
Feature is completed, ready for review 🙂

src/algoliasearch.helper.js
```diff
@@ -1201,7 +1212,13 @@ AlgoliaSearchHelper.prototype._search = function() {
 
   this._currentNbQueries++;
 
-  this.client.search(queries, this._dispatchAlgoliaResponse.bind(this, states, queryId));
+  try {
+    this.client.search(queries)
+      .then(this._dispatchAlgoliaResponse.bind(this, states, queryId))
+      .catch(this._dispatchAlgoliaError.bind(this, queryId));
+  } catch (err) {
+    this.emit('error', err);
```
bobylito: You could probably add a comment here explaining that if we reach this part, we're in a internal error state.

test/spec/algoliasearch.helper/pendingSearch.js
bobylito: SearchOnce should continue to accept callbacks, even if we've changed the internal implementation. Therefore  we should not remove this test, I think.

test/spec/algoliasearch.helper/queryID.js
```diff
@@ -47,13 +48,13 @@ test('the response handler should check that the query is not outdated', functio
     helper: helper
   }];
 
-  helper._dispatchAlgoliaResponse(states, helper._lastQueryIdReceived + 1, null, testData.response);
-  helper._dispatchAlgoliaResponse(states, helper._lastQueryIdReceived + 10, null, testData.response);
+  helper._dispatchAlgoliaResponse(states, helper._lastQueryIdReceived + 1, testData.response);
```
bobylito: There is a test missing (even before this PR) with the behavior of the outdated query when the search returns errors. Since now the implementation is split into two parts, it's more obvious. We should test it.

francoischalifour: Gotcha!

test/spec/algoliasearch.helper/pendingSearch.js
francoischalifour: Yep, sorry I got confused!

francoischalifour:
Everything has been updated @bobylito.

PR #574, opened 2018-04-20
fix(sffv): unwrap content when it comes from multi queries

### LINKED ISSUE #573: client.searchForFacetValues & index.searchForFacetValues are not interoperable
**Issue:**

The recent change that switch on `client.searchForFacetValues` when available (see #572) breaks the signature of the `searchForFacetValues` of the helper. Because the function on the `index` perform only a single query but the client method perform a multi query. The return type is different between the functions.

- with `index`: `(facet: string, query: string, ...) => Promise<FacetSearchResult>`
- with `client`: `(facet: string, query: string, ...) => Promise<Array<FacetSearchResult>>`

This issue breaks the React InstantSearch implementation (see https://github.com/algolia/react-instantsearch/issues/1177).

**Solution:**

Unwrap the array when `searchForFacetValues` return one.

samouss:
Fix #573 

_The CI should fail since locally a test is broken on a fresh install._

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -11,6 +11,7 @@ var events = require('events');
 var flatten = require('lodash/flatten');
 var forEach = require('lodash/forEach');
 var isEmpty = require('lodash/isEmpty');
+var isArray = require('lodash/isArray');
 var map = require('lodash/map');
 
 var url = require('./url');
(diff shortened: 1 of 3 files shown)


src/algoliasearch.helper.js
```diff
@@ -294,6 +295,9 @@ AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxF
   return searchForFacetValuesPromise.then(function addIsRefined(content) {
     self._currentNbQueries--;
     if (self._currentNbQueries === 0) self.emit('searchQueueEmpty');
+
+    content = isArray(content) ? content[0] : content;
```
Haroenv: Array.isArray is sufficiently supported (IE8)

samouss: It's IE9 but still ok to use it, since we support IE9+ with the helper.

Haroenv: Fine here, but I saw there are some other usages of it, the lodash/isArray only does `module.exports = Array.isArray` anyway

PR #575, opened 2018-04-20
refactor(lodash): don't use lodash/isarray

Haroenv:
This function is deprecated, and all it exports is Array.isArray anyway.

Won't have a significant impact on bundle size, but it's at least less to think about :)

Array.isArray is supported in IE9+ as we do with the rest of the helper (and we already used it in some places)

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -10,7 +10,6 @@ var reduce = require('lodash/reduce');
 var omit = require('lodash/omit');
 var indexOf = require('lodash/indexOf');
 var isNaN = require('lodash/isNaN');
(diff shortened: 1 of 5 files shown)


src/SearchParameters/index.js
```diff
@@ -90,7 +89,9 @@ function findArray(array, searchedValue) {
 }
  */
 function SearchParameters(newParameters) {
-  var params = newParameters ? SearchParameters._parseNumbers(newParameters) : {};
+  var params = newParameters
```
Haroenv: ugh I accidentally formatted the code base, let's undo those quickly

PR #576, opened 2018-04-20
chore(lodash): remove isNaN

Haroenv:
_.isNaN behaves like Number.isNaN according to the docs: https://lodash.com/docs#isNaN

and Number.isNaN is supported where we want it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

Also a really small impact, but quickly saw it while researching the isArray situation

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -9,7 +9,6 @@ var map = require('lodash/map');
 var reduce = require('lodash/reduce');
 var omit = require('lodash/omit');
 var indexOf = require('lodash/indexOf');
-var isNaN = require('lodash/isNaN');
 var isArray = require('lodash/isArray');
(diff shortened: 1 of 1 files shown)


samouss:
`Number.isNaN` is not supported on IE.

bobylito:
> Number.isNaN is not supported on IE.

but the support can be provided by polyfill.io https://polyfill.io/v2/docs/features/

Haroenv:
oh, I misread the support table and saw the `yes` for edge..., package is really small so I'm not sure if it's worth for now

bobylito:
> oh, I misread the support table and saw the yes for edge..., package is really small so I'm not sure if it's worth for now

Ok let's close for now. Thanks for taking the time.

PR #577, opened 2018-04-20
feat: Make `addAlgoliaAgent()` and `clearCache()` optional

francoischalifour:
To simplify the usage of future search clients, a good solution is to make `addAlgoliaAgent()` and `clearCache()` optional. This PR also update the now unnecessary `addAlgoliaAgent` option to fake clients in the tests.


francoischalifour:
@Haroenv This test should already cover this case: https://github.com/algolia/algoliasearch-helper-js/blob/d12cbda2bb8ebafdf3d5f9e442378d3efb7353ee/test/spec/algoliasearch.helper/client.js#L15-L24

test/spec/algoliasearch.helper/instantiate.js
```diff
@@ -1,14 +0,0 @@
-'use strict';
-
-var test = require('tape');
-var algoliasearchHelper = require('../../../index.js');
-
-test('It should instantiate even with an old client that does not support addAlgoliaAgent', function(t) {
```
bobylito: It should still pass actually :)

src/algoliasearch.helper.js
```diff
@@ -119,8 +119,9 @@ var version = require('./version');
  * just an object containing the properties you need from it.
  */
 function AlgoliaSearchHelper(client, index, options) {
-  if (!client.addAlgoliaAgent) console.log('Please upgrade to the newest version of the JS Client.'); // eslint-disable-line
-  else if (!doesClientAgentContainsHelper(client)) client.addAlgoliaAgent('JS Helper ' + version);
+  if (client.addAlgoliaAgent && !doesClientAgentContainsHelper(client)) {
```
bobylito: We don't explicitly test for addAlgoliaAgent usage with a client with  and without the method.

src/algoliasearch.helper.js
```diff
@@ -1305,7 +1306,7 @@ AlgoliaSearchHelper.prototype._change = function(newState) {
  * @return {AlgoliaSearchHelper}
  */
 AlgoliaSearchHelper.prototype.clearCache = function() {
-  this.client.clearCache();
+  this.client.clearCache && this.client.clearCache();
```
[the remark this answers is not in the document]
francoischalifour: I added the tests.

src/algoliasearch.helper.js
francoischalifour: I added the tests.

test/spec/algoliasearch.helper/client.js
```diff
@@ -12,6 +12,56 @@ function makeFakeClient() {
   return client;
 }
 
+test("client without addAlgoliaAgent() doesn't throw on instanciation", function(t) {
```
Haroenv: nit: instantiation

francoischalifour: Oops

src/algoliasearch.helper.js
Haroenv: What about exposing `addAlgoliaAgent` like we do for `clearCache`, that way we can assume in InstantSearch implementations that the method exists and just call it, even if it doesn't on the client

PR #604, opened 2018-05-07
doc(params): Add filters to query parameters

### LINKED ISSUE #528: Documentation doesn't list "filters" as a parameter for method "setQueryParameter"
I had a moderately complex numeric filter requirement and couldn't find any way to do it with the search helper. After contacting Algolia technical support, I was informed of `searchHelper.setQueryParameter('filters', '...')`

I know it's in the API documentation (which is becoming increasingly difficult to find with the new site layout) but it's very difficult to connect the dots when the documentation makes it seems as though everything that can be done is listed.

julienbourdeau:
fix #528 

In case this table was autogenerated from somewhere else, feel free to close this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1012,6 +1012,18 @@ helper.setQueryParameter('hitsPerPage', 20).search();
     </tr>
     <tr>
       <td valign="top" class="td4">
+        <p class="p2"><span class="s2">filters</span></p>
+      </td>
+      <td valign="top" class="td5">
+        <p class="p3"><span class="s1">string</span></p>
+      </td>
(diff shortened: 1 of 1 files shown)


Haroenv:
I'd rather remove the whole table and link the doc instead. WDYT @bobylito?

julienbourdeau:
Agreed 💯

bobylito:
I think it's the best option too @haroenv 👍👍

bobylito:
We might want to update the website too actually. But as far as this pr is concerned, removing the table and providing the link is enough, no?

PR #621, opened 2018-09-21
docs(instantiate): use existing signature of toggleRefine

Haroenv:
The way this was documented throws an error

see https://codesandbox.io/s/o7k7w28n86

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,9 +20,10 @@ var SearchResults = require('./src/SearchResults');
  * helper.on('result', function(result) {
  *   console.log(result);
  * });
- * helper.toggleRefine('Movies & TV Shows')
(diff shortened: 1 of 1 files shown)


samouss:
Yes it's because it's deprecated we should avoid to document this usage.

https://github.com/algolia/algoliasearch-helper-js/blob/fa4a93d94537f1416564408a972556cf43478958/src/algoliasearch.helper.js#L656-L660

francoischalifour:
@samouss Here we're even talking about `toggeRefine()` (not `toggleRefinement()`).

Haroenv:
good point, let's move this to a one that's actually correct

samouss:
I've updated the code, it's still deprecated.

PR #625, opened 2018-10-08
chore(deps): update st

Haroenv:
Also added the integrity field to the whole yarn.lock (couldn't prevent this, but did it in two commits)

vulnerability fixed: https://nvd.nist.gov/vuln/detail/CVE-2017-16224


Haroenv:
Yarn isn't enforced in the Helper, so I can't enforce a different version, but it will be clear we don't want this to be removed :)

PR #637, opened 2018-12-21
chore: use only yarn (no more npm 👩‍🚒)

Haroenv:


.travis.yml
```diff
@@ -1,11 +1,14 @@
 language: node_js
-before_install: npm prune
-before_cache: npm prune
-install: npm install
+before_install: 
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.13.0
+  - export PATH=$HOME/.yarn/bin:$PATH
+cache:
+  yarn: true
+install: yarn
```
samouss: You don't need this, if `yarn` is available it's used by default.

PR #647, opened 2019-03-22
fix(ua): change the User-Agent to use the new specs lib (version)

robertmogos:
change the User-Agent to use the new specs lib (version) from `JS Helper $version` to `JS Helper (${version})`

### DIFF
--- a/dist/algoliasearch.helper.js
+++ b/dist/algoliasearch.helper.js
@@ -13727,7 +13727,7 @@ var version = require('./version');
  */
 function AlgoliaSearchHelper(client, index, options) {
   if (client.addAlgoliaAgent && !doesClientAgentContainsHelper(client)) {
-    client.addAlgoliaAgent('JS Helper ' + version);
+    client.addAlgoliaAgent('JS Helper (' + version + ')');
   }
 
   this.setClient(client);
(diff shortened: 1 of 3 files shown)


Haroenv:
(shouldn't be merged before it's clarified we can deal with the new format)

dist/algoliasearch.helper.js
```diff
@@ -13727,7 +13727,7 @@ var version = require('./version');
  */
 function AlgoliaSearchHelper(client, index, options) {
   if (client.addAlgoliaAgent && !doesClientAgentContainsHelper(client)) {
-    client.addAlgoliaAgent('JS Helper ' + version);
+    client.addAlgoliaAgent('JS Helper (' + version + ')');
```
francoischalifour: The `dist/` files are generated, they're not supposed to be edited manually. You can revert these modifications.

src/algoliasearch.helper.js
```diff
@@ -1331,7 +1331,7 @@ AlgoliaSearchHelper.prototype.clearCache = function() {
 AlgoliaSearchHelper.prototype.setClient = function(newClient) {
   if (this.client === newClient) return this;
 
-  if (newClient.addAlgoliaAgent && !doesClientAgentContainsHelper(newClient)) newClient.addAlgoliaAgent('JS Helper ' + version);
+  if (newClient.addAlgoliaAgent && !doesClientAgentContainsHelper(newClient)) newClient.addAlgoliaAgent('JS Helper (' + version + ')');
```
francoischalifour: Can we avoid inlining `if` statements?

robertmogos: I can change this but the entire lib is using this. Does it make sense to just change it in one place?

dist/algoliasearch.helper.js
robertmogos: Done

src/algoliasearch.helper.js
francoischalifour: As edited in [the first UA change](https://github.com/algolia/algoliasearch-helper-js/pull/647/files#diff-cbf795d639d5f4b610b6d0f6ee395b2aR123), we don't use an inline condition. We should respect that rule for edited code IMO.

robertmogos: Done

PR #648, opened 2019-03-22
fix(results): remove lodash looping over objects

### LINKED ISSUE #639: Avoid to rely on Lodash to loop over collections
> Related: https://github.com/algolia/algoliasearch-helper-js/issues/552

The Helper uses Lodash internally to loop over collections. We should avoid to rely on it because we don’t own the data, users does. We can have issues when the collections contain data that Lodash uses to detect the kind of the collection e.g. an object with the key `length`. We do have the issues for the `SearchResults` but it might be present in other spots.

https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/SearchResults/index.js#L548-L576

Haroenv:
Closes #639 

This was a quite complicated effort, because most iterations could be of either type.

Migrations done: 

- remove lodash.map
- remove lodash.get (just a small side effect)
- remove map / reduce / trim / pickBy in generateHierarchicalTree
- add some more typing via jsDoc (typescript reads it), but not everywhere
- went through every usage for every lodash function in SearchResults
- did some small moves here and there to remove a function's usage fully

This fixes a bug when a facet value can be `length`, causing lodash to interpret it as an array


src/SearchResults/index.js
```diff
@@ -740,42 +782,63 @@ SearchResults.prototype.getRefinements = function() {
...
   var facet = find(resultsFacets, {name: attributeName});
-  var count = get(facet, 'data[' + name + ']');
-  var exhaustive = get(facet, 'exhaustive');
   return {
     type: type,
     attributeName: attributeName,
     name: name,
-    count: count || 0,
-    exhaustive: exhaustive || false
+    count: (facet && facet.data[name]) || 0,
```
francoischalifour: I'm not sure if `data` is necessarily part of `facet` at this point. If it is, ignore the rest of this comment.

The exact equivalent is:

```suggestion
    count: (facet && facet.data && facet.data[name]) || 0,
```

I tried [Lodash `get` REPL](https://lodash.com/docs/4.17.11#get):

```js
const facetName = 'name'
const facet = {}

const res1 = _.get(facet, 'data[facetName]')
// => undefined

const res2 = facet && facet.data[facetName]
// => TypeError: Cannot read property 'name' of undefined

const res3 = facet && facet.data && facet.data[facetName]
// => undefined
```

src/SearchResults/index.js
```diff
@@ -740,42 +782,63 @@ SearchResults.prototype.getRefinements = function() {
...
   var facetDeclaration = state.getHierarchicalFacetByName(attributeName);
-  var splitted = name.split(facetDeclaration.separator);
-  var configuredName = splitted[splitted.length - 1];
-  for (var i = 0; facet !== undefined && i < splitted.length; ++i) {
-    facet = find(facet.data, {name: splitted[i]});
-  }
-  var count = get(facet, 'count');
-  var exhaustive = get(facet, 'exhaustive');
+  var separator = state._getHierarchicalFacetSeparator(facetDeclaration);
+  var split = name.split(separator);
```
francoischalifour: Renaming the `split` variable would be slightly less confusing since there's a `split` method.

src/SearchResults/index.js
```diff
@@ -551,23 +581,23 @@ function extractNormalizedFacetValues(results, attribute) {
     var facet = find(results.facets, predicate);
     if (!facet) return [];
 
-    return map(facet.data, function(v, k) {
+    return Object.keys(facet.data).map(function(name) {
```
francoischalifour: Is `facet.data` always defined?

src/SearchResults/index.js
Haroenv: I think it might always be available, but let's be safe here indeed

src/SearchResults/index.js
Haroenv: it used to be called `splitted`, but that's grammatically incorrect. Do you have a better suggestion?

francoischalifour: Something like `hierarchicalParts`?

src/SearchResults/index.js
francoischalifour: (if not, there are some other places to update too)

Haroenv:
From reading https://github.com/algolia/algoliasearch-helper-js/blob/e111b8558c3a48a17369b03d22a00bd24cfd7e6c/src/SearchResults/index.js#L363-L538, it's always available

src/SearchResults/index.js
```diff
@@ -79,13 +80,25 @@ function assignFacetStats(dest, facetStats, key) {
...
+ * @param {string} hierarchicalAttributeName
+ */
+function findMatchingHierarchicalFacetFromAttributeName(
+  hierarchicalFacets,
+  hierarchicalAttributeName
+) {
+  return find(hierarchicalFacets, function facetKeyMatchesAttribute(
+    hierarchicalFacet
+  ) {
+    return includes(hierarchicalFacet.attributes, hierarchicalAttributeName);
```
tkrugg: This maybe can be replaced with 
```js
hierarchicalFacet.attributes.indexOf(hierarchicalAttributeName) > -1
```
?

tkrugg: looks like lodash would also check `hierarchicalFacet.attributes !== null`
https://github.com/lodash/lodash/blob/4.17.11/lodash.js#L598

Haroenv: Good call, I didn't replace that one since it's an array, but let's do it too

Haroenv:
removing lodash.includes brought it from 35.2 → 35 kB

src/SearchResults/index.js
```diff
@@ -584,7 +614,8 @@ function recSort(sortFn, node) {
   if (!node.data || node.data.length === 0) {
     return node;
   }
-  var children = map(node.data, partial(recSort, sortFn));
+
+  var children = node.data.map(partial(recSort, sortFn));
```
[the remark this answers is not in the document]
Haroenv: Yes, but it didn't have a significant impact. Note that this PR is mainly for removing the looping in lodash, not _all_ lodash (that's a separate work)

src/SearchResults/index.js
samouss: @Haroenv You did the change already?

src/SearchResults/index.js
```diff
@@ -740,42 +782,63 @@ SearchResults.prototype.getRefinements = function() {
...
-    facet = find(facet.data, {name: splitted[i]});
-  }
-  var count = get(facet, 'count');
-  var exhaustive = get(facet, 'exhaustive');
+  var separator = state._getHierarchicalFacetSeparator(facetDeclaration);
+  var split = name.split(separator);
+  var configuredName = split[split.length - 1];
+  var rootFacet = find(resultsFacets, {name: attributeName});
+
+  var facet = split.reduce(function(intermediateFacet, part) {
```
samouss: I'm not exactly sure the two are equivalent. In the previous version if the result of the `find` is `undefined` we break the loop. Which means that the rest of the function is executed with `facet = undefined`. But in the new version **we don't** break the loop. The result of this is that the last facet found is returned and the rest of the function is executed with `facet = lastFacetFound`. WDYT?

test/spec/SearchResults/getRefinements.js
```diff
@@ -107,11 +107,15 @@ test(
 
     var refinements = result.getRefinements();
 
-    t.deepEqual(refinements, [{
-      attributeName: 'hierarchicalCategories', count: 0, exhaustive: false,
-      name: 'Best Buy Gift Cards > Entertainment Gift Cards', type: 'hierarchical'
-    }]);
-
+    t.deepEqual(refinements, [
```
samouss: Why the assertions change?

Haroenv: because I fixed a bug in getRefinements with the reduce, as you can see in the JSON, there are supposed to be 17 hits, not 0

PR #653, opened 2019-04-08
refactor(events): replace util.inherits by inline

Haroenv:
I've experimented with replacing `events` by `mitt`, and the results were promising, but I think either our tests or something else still relied on the way events are flushed by the real & node module. This is why I think a more useful refactor for the events package is to remove events all together (not easily feasible without breaking change)

This is not a breaking change!

bundle impact: 35.3 kB → 32.4 kB


src/functions/inherits.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+function inherits(ctor, superCtor) {
+  ctor.prototype = Object.create(superCtor.prototype, {
```
samouss: Sure the two are absolutely equivalent? I would have replace it with the implementation of [`util.inherits`](https://github.com/nodejs/node/blob/29867f35d89c9cc0ef1ea7dfdecd6361dc00a7c1/lib/util.js#L133-L151) (minus the error validation).

Haroenv: Looks like the node implementation changed, however I used the one from `inherits` (which is used in the `util` npm package): 

https://github.com/isaacs/inherits/blob/f721d6bee2d6df13262a190fee1f48e21ed72b42/inherits_browser.js#L3-L13

Haroenv: that's likely because Object.setPrototypeOf is IE11+ only

PR #655, opened 2019-04-23
refactor(lodash): remove _.omit

Haroenv:
replaced by the loose babel helper

### DIFF
--- a/src/functions/omit.js
+++ b/src/functions/omit.js
@@ -1,0 +1,18 @@
+'use strict';
+
+// https://github.com/babel/babel/blob/3aaafae053fa75febb3aa45d45b6f00646e30ba4/packages/babel-helpers/src/helpers.js#L604-L620
+function _objectWithoutPropertiesLoose(source, excluded) {
+  if (source === null) return {};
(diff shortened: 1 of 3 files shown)


Haroenv:
2kb impact on bundle 😏

src/functions/omit.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+// https://github.com/babel/babel/blob/3aaafae053fa75febb3aa45d45b6f00646e30ba4/packages/babel-helpers/src/helpers.js#L604-L620
+function _objectWithoutPropertiesLoose(source, excluded) {
+  if (source === null) return {};
+  var target = {};
+  var sourceKeys = Object.keys(source);
```
samouss: I think we might have the same issue than InstantSearch under IE. We have to be sure that the `source` is not `undefined` otherwise it might throw an error.

Haroenv: both calls to `omit` are guarded by isEmpty (and changed to === undefined in #656, since they can only be undefined or something)

PR #656, opened 2019-04-23
refactor(lodash): remove simple functions

Haroenv:
- isUndefined
- isEqual
- isNaN
- keys
- isFunction
- isEmpty

### DIFF
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -90,7 +90,7 @@ var lib = {
    */
   clearRefinement: function clearRefinement(refinementList, attribute, refinementType) {
     if (attribute === undefined) {
-      if (!(refinementList && objectHasKeys(refinementList))) {
+      if (!objectHasKeys(refinementList)) {
         return refinementList;
       }
       return {};
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -13,7 +13,6 @@
  */
 
 var isUndefined = require('lodash/isUndefined');
-var isFunction = require('lodash/isFunction');
 var isEmpty = require('lodash/isEmpty');
 
 var defaultsPure = require('../functions/defaultsPure');
@@ -98,7 +97,7 @@ var lib = {
     } else if (typeof attribute === 'string') {
       if (isEmpty(refinementList[attribute])) return refinementList;
       return omit(refinementList, attribute);
-    } else if (isFunction(attribute)) {
+    } else if (typeof attribute === 'function') {
       var hasChanged = false;
 
       var newRefinementList = Object.keys(refinementList).reduce(function(memo, key) {
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -12,9 +12,6 @@
  * @typedef {Object.<string, SearchParameters.refinementList.Refinements>} SearchParameters.refinementList.RefinementList
  */
 
-var isUndefined = require('lodash/isUndefined');
-var isEmpty = require('lodash/isEmpty');
-
 var defaultsPure = require('../functions/defaultsPure');
 var omit = require('../functions/omit');
 
@@ -53,7 +50,7 @@ var lib = {
    * @return {RefinementList} a new and updated refinement lst
    */
   removeRefinement: function removeRefinement(refinementList, attribute, value) {
-    if (isUndefined(value)) {
+    if (value === undefined) {
       return lib.clearRefinement(refinementList, attribute);
     }
 
@@ -71,7 +68,7 @@ var lib = {
    * @return {RefinementList} a new and updated list
    */
   toggleRefinement: function toggleRefinement(refinementList, attribute, value) {
-    if (isUndefined(value)) throw new Error('toggleRefinement should be used with a value');
+    if (value === undefined) throw new Error('toggleRefinement should be used with a value');
 
     if (lib.isRefined(refinementList, attribute, value)) {
       return lib.removeRefinement(refinementList, attribute, value);
@@ -91,11 +88,11 @@ var lib = {
    * @return {RefinementList} a new and updated refinement list
    */
   clearRefinement: function clearRefinement(refinementList, attribute, refinementType) {
-    if (isUndefined(attribute)) {
-      if (isEmpty(refinementList)) return refinementList;
+    if (attribute === undefined) {
+      if (Object.keys(refinementList).length === 0) return refinementList;
       return {};
     } else if (typeof attribute === 'string') {
-      if (isEmpty(refinementList[attribute])) return refinementList;
+      if (refinementList[attribute] === undefined) return refinementList;
       return omit(refinementList, attribute);
     } else if (typeof attribute === 'function') {
       var hasChanged = false;
@@ -105,12 +102,14 @@ var lib = {
         var facetList = values.filter(function(value) {
           return !attribute(value, key, refinementType);
         });
-
-        if (!isEmpty(facetList)) {
-          if (facetList.length !== values.length) hasChanged = true;
+        if (facetList.length > 0) {
+          if (facetList.length !== values.length) {
+            hasChanged = true;
+          }
           memo[key] = facetList;
+        } else {
+          hasChanged = true;
         }
-        else hasChanged = true;
 
         return memo;
       }, {});
@@ -132,7 +131,7 @@ var lib = {
     var containsRefinements = !!refinementList[attribute] &&
       refinementList[attribute].length > 0;
 
-    if (isUndefined(refinementValue) || !containsRefinements) {
+    if (refinementValue === undefined || !containsRefinements) {
       return containsRefinements;
     }
 
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -14,6 +14,7 @@
 
 var defaultsPure = require('../functions/defaultsPure');
 var omit = require('../functions/omit');
+var objectHasKeys = require('../functions/objectHasKeys');
 
 var lib = {
   /**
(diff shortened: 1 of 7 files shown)


src/SearchParameters/RefinementList.js
```diff
@@ -94,25 +91,27 @@ var lib = {
...
    */
   clearRefinement: function clearRefinement(refinementList, attribute, refinementType) {
-    if (isUndefined(attribute)) {
-      if (isEmpty(refinementList)) return refinementList;
+    if (attribute === undefined) {
+      if (Object.keys(refinementList).length === 0) return refinementList;
       return {};
     } else if (typeof attribute === 'string') {
-      if (isEmpty(refinementList[attribute])) return refinementList;
+      if (refinementList[attribute] === undefined) return refinementList;
```
francoischalifour: This is not similar. Is that expected?

src/SearchParameters/index.js
```diff
@@ -596,14 +612,19 @@ SearchParameters.validate = function(currentState, parameters) {
       'an error, if it is not, you should first clear the tags with clearTags method.');
   }
 
-  if (currentState.numericFilters && params.numericRefinements && !isEmpty(params.numericRefinements)) {
+  if (
+    currentState.numericFilters &&
+    params.numericRefinements &&
+    !objectHasKeys(params.numericRefinements)
```
francoischalifour: The condition should be reversed.

```suggestion
    objectHasKeys(params.numericRefinements)
```

src/SearchParameters/index.js
```diff
@@ -506,6 +521,7 @@ SearchParameters._parseNumbers = function(partialState) {
     var value = partialState[k];
     if (typeof value === 'string') {
       var parsedValue = parseFloat(value);
+      // global isNaN i sok to use here, value is only number or NaN
```
francoischalifour:
```suggestion
      // global isNaN is ok to use here, value is only number or NaN
```

src/SearchParameters/index.js
```diff
@@ -893,13 +920,17 @@ SearchParameters.prototype = {
...
-    if (isUndefined(attribute)) {
-      if (isEmpty(this.numericRefinements)) return this.numericRefinements;
+    if (attribute === undefined) {
+      if (!objectHasKeys(this.numericRefinements)) {
+        return this.numericRefinements;
+      }
       return {};
     } else if (typeof attribute === 'string') {
-      if (isEmpty(this.numericRefinements[attribute])) return this.numericRefinements;
+      if (this.numericRefinements[attribute] === undefined) {
```
francoischalifour: Is that the same condition?

src/SearchParameters/index.js
```diff
@@ -1593,7 +1629,7 @@ SearchParameters.prototype = {
     var parsedParams = SearchParameters._parseNumbers(params);
 
     return this.mutateMe(function mergeWith(newInstance) {
-      var ks = keys(params);
+      var ks = Object.keys(params);
```
francoischalifour: Since global `keys` doesn't exist anymore, we can rename this variable to `keys`.

src/algoliasearch.helper.js
```diff
@@ -118,7 +117,7 @@ var version = require('./version');
  * just an object containing the properties you need from it.
  */
 function AlgoliaSearchHelper(client, index, options) {
-  if (client.addAlgoliaAgent && !doesClientAgentContainsHelper(client)) {
```
francoischalifour: Should this be part of this PR?

src/algoliasearch.helper.js
```diff
@@ -1330,7 +1329,7 @@ AlgoliaSearchHelper.prototype.clearCache = function() {
 AlgoliaSearchHelper.prototype.setClient = function(newClient) {
   if (this.client === newClient) return this;
 
-  if (newClient.addAlgoliaAgent && !doesClientAgentContainsHelper(newClient)) {
```
francoischalifour: Should this be part of this PR?

src/SearchParameters/RefinementList.js
Haroenv: isEmpty will also be true for undefined

src/SearchParameters/index.js
```diff
@@ -25,17 +19,38 @@ var filterState = require('./filterState');
 
 var RefinementList = require('./RefinementList');
 
+function objectHasKeys(obj) {
+  return Object.keys(obj).length > 0;
+}
```
yannickcr: IMHO it could be moved to an external file in `functions` so it could be used in multiple files without the need to duplicate it. Same for `isEqualNumArr`.

src/SearchParameters/index.js
```diff
@@ -25,17 +19,38 @@ var filterState = require('./filterState');
...
+
+/**
+ * isEqual, but only for arrays of base elements, or base elements
+ * @param {number|Array<number>} a first element
+ * @param {number|Array<number>} b second element
+ */
+function isEqualNumArr(a, b) {
```
Haroenv: this one is not exactly reusable (on purpose), because it's only used in findArray which is only used in one spot. I preferred here to just make the existing structure work again with isEqual removed, but the better solution is to get rid of both helpers I think. That can be done at a later point?

src/SearchParameters/RefinementList.js
Haroenv: according to the tests at least, it's either present in the object, either undefined.

src/SearchParameters/index.js
yannickcr: Yes, it could totally be done at a later point :+1:

src/SearchParameters/filterState.js
```diff
@@ -28,20 +27,20 @@ function filterState(state, filters) {
       }
 
       var numericRefinements = state.getNumericRefinements(attr);
-      if (!isEmpty(numericRefinements)) {
+      if (objectHasKeys(numericRefinements)) {
```
tkrugg: I think getNumericRefinement might be returning arrays an not objects. [see test](https://github.com/algolia/algoliasearch-helper-js/blob/633c24190fb7d23d724ff6ac990116317d885dcd/test/spec/algoliasearch.helper/getNumericRefinement.js#L21)
https://github.com/algolia/algoliasearch-helper-js/blob/633c24190fb7d23d724ff6ac990116317d885dcd/test/spec/algoliasearch.helper/getNumericRefinement.js#L21

src/algoliasearch.helper.js
```diff
@@ -959,7 +958,7 @@ AlgoliaSearchHelper.prototype.isRefined = function(facet, value) {
  *
  */
 AlgoliaSearchHelper.prototype.hasRefinements = function(attribute) {
-  if (!isEmpty(this.state.getNumericRefinements(attribute))) {
+  if (Object.keys(this.state.getNumericRefinements(attribute)).length > 0) {
```
tkrugg: this.state.getNumericRefinements(attribute) seems to be an array
Redundant, but adding it since it's not using `objectHasKeys` like the rest.

src/SearchParameters/filterState.js
Haroenv: You're looking at it with two arguments, with one argument it does actually return an object

src/algoliasearch.helper.js
Haroenv: looks like it was forgotten to be refactored

src/SearchParameters/index.js
```diff
@@ -1,36 +1,48 @@
...
 
 var RefinementList = require('./RefinementList');
 
 /**
- * like _.find but using _.isEqual to be able to use it
+ * isEqual, but only for arrays of base elements, or base elements
+ * @param {number|Array<number>} a first element
+ * @param {number|Array<number>} b second element
+ */
+function isEqualNumArr(a, b) {
```
tkrugg: We call it `isEqual*Num*Array` but it actually works with all primitive types like the doc you added says it.
There might be a consistency in the naming there.
I don't have a better naming suggestion 🐵 maybe `isShallowEqualArray`

francoischalifour: We could make this function work only for arrays, call it something like `isArrayEqual` and leave `a === b` apart. Since this is just an identity, we can use the identity on top of `isArrayEqual` whenever the old `isEqual` was used.

src/SearchParameters/index.js
```diff
@@ -1454,21 +1489,22 @@ SearchParameters.prototype = {
...
+    if (value === undefined || !isOperatorDefined) {
       return isOperatorDefined;
     }
 
     var parsedValue = valToNumber(value);
-    var isAttributeValueDefined = !isUndefined(
-      findArray(this.numericRefinements[attribute][operator], parsedValue)
-    );
+    var isAttributeValueDefined =
+      findArray(this.numericRefinements[attribute][operator], parsedValue) !==
```
Haroenv: This should be replaced by a regular find

src/SearchParameters/index.js
Haroenv: called it `isEqualNumericRefinement`

src/SearchParameters/filterState.js
Haroenv: was already covered by existing tests :)

src/SearchParameters/index.js
```diff
@@ -565,11 +584,16 @@ SearchParameters.prototype = {
    */
   removeNumericRefinement: function(attribute, operator, paramValue) {
     if (paramValue !== undefined) {
-      var paramValueAsNumber = valToNumber(paramValue);
```
samouss: Is it expected that we dropped this conversion?

src/SearchParameters/index.js
```diff
@@ -565,11 +584,16 @@ SearchParameters.prototype = {
...
+      if (!this.isNumericRefined(attribute, operator, paramValue)) {
+        return this;
+      }
       return this.setQueryParameters({
         numericRefinements: this._clearNumericRefinements(function(value, key) {
-          return key === attribute && value.op === operator && isEqual(value.val, paramValueAsNumber);
+          return (
+            key === attribute &&
+            value.op === operator &&
+            isEqualNumericRefinement(value.val, valToNumber(paramValue))
```
samouss: We've inlined the conversion of the value but it was used inside the condition.

src/functions/objectHasKeys.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+function objectHasKeys(obj) {
+  return Object.keys(obj).length > 0;
```
samouss: We could support `undefined` to avoid the extra check all over the place.

src/SearchParameters/index.js
Haroenv: yes, isNumericRefined already calls `valToNumber` on the value: https://github.com/algolia/algoliasearch-helper-js/pull/656/files#diff-ecd72976141723d9448df45e667cb4beR1224

src/SearchParameters/index.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/656/files#r282869915

PR #657, opened 2019-04-23
refactor(addAgent): remove duplicate code

Haroenv:
could be considered as a breaking change if you're using a client version < 3.20.4 https://github.com/algolia/algoliasearch-client-javascript/commit/77b97629dbc692401c07b0d03854d6a93c57022c (and > 3.8.0, when addAlgoliaAgent was added https://github.com/algolia/algoliasearch-client-javascript/commit/3370fa48bc5dc66a84769dd19160be47597eb21e)

In the case of the older client being used, the agent will simply be injected multiple times, as was before that fix was added

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -118,7 +118,7 @@ var version = require('./version');
  * just an object containing the properties you need from it.
  */
 function AlgoliaSearchHelper(client, index, options) {
-  if (client.addAlgoliaAgent) {
+  if (typeof client.addAlgoliaAgent === 'function') {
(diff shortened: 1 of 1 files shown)


src/algoliasearch.helper.js
```diff
@@ -118,7 +118,7 @@ var version = require('./version');
  * just an object containing the properties you need from it.
  */
 function AlgoliaSearchHelper(client, index, options) {
-  if (client.addAlgoliaAgent && !doesClientAgentContainsHelper(client)) {
+  if (client.addAlgoliaAgent) {
```
samouss: We should use `typeof ...` otherwise we might throw an error if the value is not a function.

PR #659, opened 2019-04-25
chore(Jest): setup migration

samouss:
This is the first step towards the migration to Jest. It setups the configuration/requirement for Jest to run on the project. The PRs that follow migrate the tests folder by folder to limit the amount of lines to review.

- The update to Node 8.x.x is required by the dependency [`sane`](https://github.com/amasad/sane/blob/64ff3a870c42e84f744086884bf55a4f9c22d376/package.json#L48-L50) that Jest relies on.
- The Jest CLI only runs the unit test (`test/spec`) by default. To run the unit + integration we have to use the `test` command. It eases the development workflow to only runs the unit test from the CLI.
- We use a super old version of ESLint. The `jest` env used on this version does not include the `test` identifier, that's why I've enabled it. Only the `test` folder is impacted by the change.


.nvmrc
```diff
@@ -1 +1 @@
-7.10.0
+8.9.0
```
Haroenv: have you tried node 10? Node 8 is EOL soon

samouss: I picked `8.x.x` but `10.x.x` should also work.

samouss:
@Haroenv Yes it's expected it's only the Jest setup.

PR #666, opened 2019-04-25
chore(Jest): runner [PART-7]

samouss:
This PR updates the test runner that was implemented with `tape`. We use the exact same behaviour than before expect that now both unit and integration tests run with `jest`. You can run the unit tests locally with:

```
yarn test
```

To run both unit and integration tests:

```
INTEGRATION_TEST_API_KEY=yourApiKey \
INTEGRATION_TEST_APPID=yourAppId \
yarn test
```

I've increased the default timeout to `15000` only when the integration tests are expected to run. By default `tape` does not timeout on async test but `jest` does. I've tested multiple times with this value it worked. We can increase it a bit more if we have flakiness issues.


Haroenv:
Travis doesn't run  here because the target branch isn't develop :(

test/run.js
```diff
@@ -1,26 +1,39 @@
...
+  if (!response.results.success) {
+    process.exitCode = response.globalConfig.testFailureExitCode;
+  }
+
+  if (enableIntegrationTest) {
+    var client = algolia(process.env.INTEGRATION_TEST_APPID, process.env.INTEGRATION_TEST_API_KEY);
+
+    client.listIndexes().then(content => {
+      content.items
+        .map(i => i.name)
```
Haroenv: these likely might need to be in regular functions? or I guess it is fine since it's only ran in node

test/run.js
```diff
@@ -1,26 +1,39 @@
 'use strict';
 
-var bulk = require('bulk-require');
-var test = require('tape');
+var path = require('path');
+var jest = require('jest');
 var algolia = require('algoliasearch');
```
francoischalifour: Usually we call this `algoliasearch`.

test/run.js
samouss: Yes indeed, it was already like that. All the tests could be wrote with modern JavaScript, but we have to update the linter to support some modern syntax like spread operator.

samouss:
> Since we use Node 8 now we could use a newer syntax.

https://github.com/algolia/algoliasearch-helper-js/pull/666#discussion_r278906039

> We can't check for sure that it works on Travis though since it's not running.

Yes indeed. Travis is set up to only runs on `develop` & `master`. It would be solved once the PRs are merged into one and targets `develop`. I've opened a PR https://github.com/algolia/algoliasearch-helper-js/pull/658 ([CI](https://travis-ci.org/algolia/algoliasearch-helper-js/builds/524408089)) to fire Travis yesterday to check that the CI passed.

PR #667, opened 2019-04-25
chore(Jest): cleanup [PART-8]

samouss:
This PR drops the support for the performance test. I've tried to increate the creation time of the Helper with (almost an infinite) loop to notice a difference. The test didn't caught it. We can safely remove it. We also drop all the dependencies required to the previous test stack `tape`, `sinon`, ...


package.json
```diff
@@ -47,17 +46,13 @@
     "babel-core": "6.26.3",
     "babel-loader": "6.4.1",
     "babel-preset-es2015": "6.24.1",
-    "bench": "0.3.6",
-    "benchmark": "2.1.4",
     "browserify": "14.5.0",
     "browzers": "1.3.0",
-    "bulk-require": "1.0.1",
```
Haroenv: I thought this was used for the other tests?

samouss: Yes maybe, which one? I didn't found any occurrences of the import anymore. I've replaced the runner implementation on https://github.com/algolia/algoliasearch-helper-js/pull/666.

Haroenv: was confused about the order of the PRs

PR #670, opened 2019-04-30
refactor(SearchParameters): removes default values

samouss:
This PR drops the support of default values inside `SearchParameters`. The current implementation keeps a list of possible values inside the class constructor. The downside of this approach is that if the value is not defined we got `undefined`. 

It wasn't an issue until now. We want to be able to easily merge a hierarchy of `SearchParameters` with the default values it's impossible to detect whether the value is omit, initial or set. This is specially true for `query` and `page`.

**With defaults:**

```json
{
  "index": "",
  "query": "",
  "facets": [],
  "disjunctiveFacets": [],
  "hierarchicalFacets": [],
  "facetsRefinements": {},
  "facetsExcludes": {},
  "disjunctiveFacetsRefinements": {},
  "numericRefinements": {},
  "tagRefinements": [],
  "hierarchicalFacetsRefinements": {},
  "numericFilters": "undefined",
  "tagFilters": "undefined",
  "optionalTagFilters": "undefined",
  "optionalFacetFilters": "undefined",
  "hitsPerPage": "undefined",
  "maxValuesPerFacet": "undefined",
  "page": 0,
  "queryType": "undefined",
  "typoTolerance": "undefined",
  "minWordSizefor1Typo": "undefined",
  "minWordSizefor2Typos": "undefined",
  "minProximity": "undefined",
  "allowTyposOnNumericTokens": "undefined",
  "ignorePlurals": "undefined",
  "restrictSearchableAttributes": "undefined",
  "advancedSyntax": "undefined",
  "analytics": "undefined",
  "analyticsTags": "undefined",
  "synonyms": "undefined",
  "replaceSynonymsInHighlight": "undefined",
  "optionalWords": "undefined",
  "removeWordsIfNoResults": "undefined",
  "attributesToRetrieve": "undefined",
  "attributesToHighlight": "undefined",
  "highlightPreTag": "undefined",
  "highlightPostTag": "undefined",
  "attributesToSnippet": "undefined",
  "getRankingInfo": "undefined",
  "distinct": "undefined",
  "aroundLatLng": "undefined",
  "aroundLatLngViaIP": "undefined",
  "aroundRadius": "undefined",
  "minimumAroundRadius": "undefined",
  "aroundPrecision": "undefined",
  "insideBoundingBox": "undefined",
  "insidePolygon": "undefined",
  "snippetEllipsisText": "undefined",
  "disableExactOnAttributes": "undefined",
  "enableExactOnSingleWordQuery": "undefined",
  "offset": "undefined",
  "length": "undefined"
}
```

**Without defaults:**

```json
{
  "facets": [],
  "disjunctiveFacets": [],
  "hierarchicalFacets": [],
  "facetsRefinements": {},
  "facetsExcludes": {},
  "disjunctiveFacetsRefinements": {},
  "numericRefinements": {},
  "tagRefinements": [],
  "hierarchicalFacetsRefinements": {}
}
```


test/spec/algoliasearch.helper/state.js
```diff
@@ -119,7 +119,11 @@ test('Get the state as a query string', function() {
   expect(algoliasearchHelper.SearchParameters._parseNumbers(decodedState)).toEqual(stateWithoutConfig);
 });
 
-test('Set the state with a query parameter with index', function() {
+// @TODO: the implementation of URL requires the `SearchParameters` to have the
+// list of attributes defined on the object. We dropped this feature, it's fine
+// because we also want to drop the URL feature from the Helper. Test tests are
```
francoischalifour:
```suggestion
// because we also want to drop the URL feature from the Helper. The tests are
```

test/spec/url.js
```diff
@@ -4,7 +4,11 @@ var algoliasearchHelper = require('../../index');
 
 var fakeClient = {};
 
-test('getStateFromQueryString should parse insideBoundingBox as float georects and be consistent with the state', function() {
+// @TODO: the implementation of URL requires the `SearchParameters` to have the
+// list of attributes defined on the object. We dropped this feature, it's fine
+// because we also want to drop the URL feature from the Helper. Test tests are
```
francoischalifour:
```suggestion
// because we also want to drop the URL feature from the Helper. The tests are
```

src/SearchParameters/index.js
```diff
@@ -89,21 +89,6 @@ function findArray(array, searchedValue) {
 function SearchParameters(newParameters) {
   var params = newParameters ? SearchParameters._parseNumbers(newParameters) : {};
 
-  /**
-   * Targeted index. This parameter is mandatory.
-   * @member {string}
-   */
-  this.index = params.index || '';
```
Haroenv: just wondering if I understand this correctly: with this removal you can no longer read the values from the parameters directly, so `new SearchParameters().index === undefined` always?

samouss: Yes you can if they are defined otherwise you get `undefined`. It changes nothing expect for `query` and `page`. `index` is required anyway so the default value was never used.

src/SearchParameters/index.js
```diff
@@ -204,264 +189,6 @@ function SearchParameters(newParameters) {
...
-   */
-  this.enableExactOnSingleWordQuery = params.enableExactOnSingleWordQuery;
-
-  // Undocumented parameters, still needed otherwise we fail
-  this.offset = params.offset;
-  this.length = params.length;
-
   var self = this;
   forOwn(params, function checkForUnknownParameter(paramValue, paramName) {
     if (SearchParameters.PARAMETERS.indexOf(paramName) === -1) {
```
Haroenv: ah I see now that the line underneath this one causes unset variables to go on `this`

samouss: Yes, exactly.

PR #673, opened 2019-05-02
feat: implement dedicated reset page method

samouss:
This PR introduces a new internal method on the `SearchParameters` called `resetPage`. This function allows differentiating a "page reset" against a "regular refinement". It wasn't an issue until now because the default value for the page was `0`. We don't have default values anymore, the value is either omit or set to a value. With the previous implementation, a call to a method that resets (implicitly) the page on a `SearchParameters` that doesn't already contain a page it creates the attribute with the value 0. This is not the expected behavior. We want to let parameter without the page.

- page omit: we don't touch the instance
- page present: we reset its value to `0`

We are now able to detect a "page reset" against a "regular refinement". The `change` event now contains an additional argument that provide the information. It allows the subscribers to trigger some specific actions based on that.

```js
const helper = algoliaSearchHelper();

helper.on("change", ({ state, results, isPageReset }) => {
  if (isPageReset) {
    // specific action
  }
});
```

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -329,7 +329,7 @@ AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxF
  * @chainable
  */
 AlgoliaSearchHelper.prototype.setQuery = function(q) {
-  this._change(this.state.setPage(0).setQuery(q));
+  this._change(this.state.resetPage().setQuery(q));
   return this;
 };
 
@@ -357,7 +357,7 @@ AlgoliaSearchHelper.prototype.setQuery = function(q) {
  * }).search();
  */
 AlgoliaSearchHelper.prototype.clearRefinements = function(name) {
-  this._change(this.state.setPage(0).clearRefinements(name));
+  this._change(this.state.resetPage().clearRefinements(name));
   return this;
 };
 
@@ -370,7 +370,7 @@ AlgoliaSearchHelper.prototype.clearRefinements = function(name) {
  * @chainable
  */
 AlgoliaSearchHelper.prototype.clearTags = function() {
-  this._change(this.state.setPage(0).clearTags());
+  this._change(this.state.resetPage().clearTags());
   return this;
 };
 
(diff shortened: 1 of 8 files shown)


test/spec/algoliasearch.helper/setQueryParameter.js
```diff
@@ -20,13 +20,7 @@ test('setChange should change the current state', function() {
 });
 
 test('setChange should not change the current state: no real modification', function() {
-  var helper = algoliasearchHelper(fakeClient, null, {
-    // @TODO: at the moment we have to provide a default value for the page, otherwise
-    // some methods reset the page to 0 (because of the reset behavior). This is what
-    // happens: page omit -> page defined with 0. We'll fix those issues with a next PR
-    // that implements a proper reset with the updated structure of the `SearchParameters`.
-    page: 0
```
Haroenv: in the other instances of this comment you removed the page 0 from the test, but you kept it here, is that expected?

src/algoliasearch.helper.js
```diff
@@ -1238,10 +1238,12 @@ AlgoliaSearchHelper.prototype._hasDisjunctiveRefinements = function(facet) {
...
 };
 
-AlgoliaSearchHelper.prototype._change = function(newState) {
+AlgoliaSearchHelper.prototype._change = function(newState, resetPage) {
+  var isPageReset = resetPage || false;
+
   if (newState !== this.state) {
     this.state = newState;
-    this.emit('change', this.state, this.lastResults);
+    this.emit('change', this.state, this.lastResults, isPageReset);
```
Haroenv: I'm not too much a fan of this boolean trap, what if we make it an options object? not sure if there's more things possible, and how long we plan on keeping the emitters though 🤔

samouss: Yes indeed, we could use an object. We already have BCs anyway we can move the arguments to an object. It would help to understand which value is provided to the listener.

We have to update the others too, it makes no sense to keep both at the same time. I'll only update the shape of the `change` event in the PR. I'll do the others (`search`, `result`, ...) in another one afterward.

test/spec/algoliasearch.helper/setQueryParameter.js
samouss: Yes it's expected. The test checks that a change does not occur between two exact same calls. The call that is used inside the test is `setPage(0)`. Now that we don't have default values anymore we have to provide a value for the initial state.

src/algoliasearch.helper.js
samouss: Resolved in https://github.com/algolia/algoliasearch-helper-js/pull/673/commits/5698b7939130b2a21fa4e20167c9c279e77440c8

PR #674, opened 2019-05-03
refactor(lodash): remove forEach

Haroenv:
I've been on the safe side and added defaults on places that probably already have defaults.

I didn't remove the `forEach` in URLs, because that file is removed in `next`

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -9,7 +9,6 @@ var events = require('events');
 var inherits = require('./functions/inherits');
 
 var flatten = require('lodash/flatten');
-var forEach = require('lodash/forEach');
 var isEmpty = require('lodash/isEmpty');
 
 var url = require('./url');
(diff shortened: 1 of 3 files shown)


src/algoliasearch.helper.js
```diff
@@ -306,10 +305,15 @@ AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxF
 
     content = Array.isArray(content) ? content[0] : content;
 
-    content.facetHits = forEach(content.facetHits, function(f) {
-      f.isRefined = isDisjunctive ?
-        state.isDisjunctiveFacetRefined(facet, f.value) :
-        state.isFacetRefined(facet, f.value);
+    content.facetHits = Array.isArray(content.facetHits)
```
samouss: For which conditions `facetHits` is not an array?

src/algoliasearch.helper.js
```diff
@@ -306,10 +305,15 @@ AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxF
...
+      ? content.facetHits
+      : [];
+
+    content.facetHits = content.facetHits.map(function(f) {
```
samouss: We should have stick with `forEach` since we perform a side effect inside the loop. The return value is useless in that case. We mutate the original ref of the value.

src/algoliasearch.helper.js
Haroenv: Some of the tests return an empty object, but indeed I don't think the API can do that 🤔

src/algoliasearch.helper.js
Haroenv: You're right, this should be forEach

src/algoliasearch.helper.js
samouss: We should fix the tests then.

Haroenv: https://github.com/algolia/algoliasearch-helper-js/pull/686

src/algoliasearch.helper.js
Haroenv: https://github.com/algolia/algoliasearch-helper-js/pull/686

PR #678, opened 2019-05-06
refactor(lodash): remove reduce

Haroenv:
I know these lines are too long, but that can be fixed another time when we have proper eslint integration


src/requestBuilder.js
```diff
@@ -250,8 +249,7 @@ var requestBuilder = {
   _getHitsHierarchicalFacetsAttributes: function(state) {
     var out = [];
 
-    return reduce(
-      state.hierarchicalFacets,
+    return state.hierarchicalFacets.reduce(
```
francoischalifour: Did you make sure that `state.hierarchicalFacets` can never be `undefined` at this point?

Haroenv: Yes, none of the members of state can be unset unless manually modified without using the functions

PR #682, opened 2019-05-06
chore(warn): remove unused function (lodash.bind)

Haroenv:
I first rewrote this function to be like the one from Vue InstantSearch, because it's the simplest one, but then saw it's never used, so removed it completely. 

Because of this there's no impact on bundle size.

(we should use a common console.warn, only once, only in dev though, seems like we're writing that multiple times)


robertmogos:
@Haroenv could you add another reviewer? I removed myself from the team

PR #686, opened 2019-05-06
test(sffv): no longer test impossible responses

Haroenv:
Since the engine will always respond with `facetHits` on searchForFacetValues, we can assume this will always be defined.

follow-up to @samouss' comments on #674

### DIFF
--- a/test/spec/algoliasearch.helper/searchForFacetValues.js
+++ b/test/spec/algoliasearch.helper/searchForFacetValues.js
@@ -2,13 +2,19 @@
 
 var algoliasearchHelper = require('../../../index');
 
+var fakeSearchForFacetValuesResponse = {
+  exhaustiveFacetsCount: true,
+  facetHits: [],
+  processingTimeMS: 3
+};
+
 test('searchForFacetValues calls the client method over the index method', function() {
   var clientSearchForFacetValues = jest.fn(function() {
-    return Promise.resolve([{}]);
+    return Promise.resolve([fakeSearchForFacetValuesResponse]);
   });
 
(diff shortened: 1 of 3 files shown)


test/spec/algoliasearch.helper/searchForFacetValues.js
```diff
@@ -2,13 +2,19 @@
 
 var algoliasearchHelper = require('../../../index');
 
+var fakeSearchForFacetValuesResponse = {
```
samouss: Since we mutate the value it's safer to use a factory to create the response.

test/spec/algoliasearch.helper/searchForFacetValues.js
```diff
@@ -2,13 +2,22 @@
...
 var algoliasearchHelper = require('../../../index');
 
+function createFakeSearchForFacetValuesResponse() {
+  return {
+    exhaustiveFacetsCount: true,
+    facetHits: [],
+    processingTimeMS: 3
+  };
+}
+
```
samouss:
```suggestion
```

test/spec/algoliasearch.helper/searchForFacetValues.js
```diff
@@ -2,13 +2,22 @@
 
 var algoliasearchHelper = require('../../../index');
 
+function createFakeSearchForFacetValuesResponse() {
```
samouss:
```suggestion
function makeFakeSearchForFacetValuesResponse() {
```

The codebase mainly uses `make` rather than `create`.

PR #687, opened 2019-05-06
refactor(lodash): replace find & findIndex

Haroenv:
These are very naive implementations which only work on arrays, but much shorter than a real polyfill

Notable difference between the previous implementation and this one is that the `find` implementation only takes in arrays, no longer objects

### DIFF
--- a/src/SearchResults/generate-hierarchical-tree.js
+++ b/src/SearchResults/generate-hierarchical-tree.js
@@ -3,8 +3,8 @@
 module.exports = generateTrees;
 
 var orderBy = require('lodash/orderBy');
-var find = require('lodash/find');
 
+var find = require('../functions/find');
 var prepareHierarchicalFacetSortBy = require('../functions/formatSort');
 
 function generateTrees(state) {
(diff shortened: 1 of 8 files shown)


francoischalifour:
We should test these functions.

Haroenv:
We wouldn't test these functions if they were inline, I don't think it needs to be tested

yannickcr:
Inline they would only have been used in a particular context, but since they are meant to be generic then I would have added dedicated tests.

src/SearchResults/generate-hierarchical-tree.js
```diff
@@ -82,7 +82,9 @@ function generateHierarchicalTree(
          * @type {object[]]} hierarchical data
          */
         var data = parent && Array.isArray(parent.data) ? parent.data : [];
-        parent = find(data, {isRefined: true});
+        parent = find(data, function(subtree) {
+          return subtree.isRefined;
```
samouss: The two are not strictly equivalent, but I guess it's fine since we should control the `isRefined` value.

PR #688, opened 2019-05-06
refactor(lodash): sumBy

Haroenv:
I added a new test file, since this didn't seem to be tested anywhere.

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -3,7 +3,6 @@
 var compact = require('lodash/compact');
 var findIndex = require('lodash/findIndex');
 
-var sumBy = require('lodash/sumBy');
 var find = require('lodash/find');
 var orderBy = require('lodash/orderBy');
 
@@ -286,7 +285,9 @@ function SearchResults(state, results) {
    * sum of the processing time of all the queries
    * @member {number}
    */
-  this.processingTimeMS = sumBy(results, 'processingTimeMS');
+  this.processingTimeMS = results.reduce(function(sum, result) {
+    return sum + result.processingTimeMS;
+  }, 0);
   /**
    * The position if the position was guessed by IP.
    * @member {string}
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -286,7 +286,9 @@ function SearchResults(state, results) {
    * @member {number}
    */
   this.processingTimeMS = results.reduce(function(sum, result) {
-    return sum + result.processingTimeMS;
+    return result.processingTimeMS === undefined
+      ? sum
+      : sum + result.processingTimeMS;
   }, 0);
   /**
    * The position if the position was guessed by IP.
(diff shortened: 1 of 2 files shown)


test/spec/SearchResults/initialization.js
```diff
@@ -0,0 +1,30 @@
...
+  var result = new SearchResults(new SearchParameters(), [
+    {
+      processingTimeMS: undefined
+    },
+    {
+      processingTimeMS: 1
+    }
+  ]);
+
+  expect(result.processingTimeMS).toBe(NaN);
```
Haroenv: this returned `1` in the previous implementation, but I don't think the engine ever returns this anyway. Added the test for this comment.

tkrugg: I think `processingTimeMS` can be `undefined` because the API says it can be filtered out: https://www.algolia.com/doc/api-reference/api-parameters/responseFields/?language=js#usage-notes

To be confirmed.

If yes, than we should at least add a check similar to the one sumBy had
https://github.com/lodash/lodash/blob/master/.internal/baseSum.js#L14

And then maybe rollback this test in that case ?

Haroenv: the previous implementation would have returned 1 there, I can also go with `=== undefined`

src/SearchResults/index.js
```diff
@@ -286,7 +285,11 @@ function SearchResults(state, results) {
    * sum of the processing time of all the queries
    * @member {number}
    */
-  this.processingTimeMS = sumBy(results, 'processingTimeMS');
+  this.processingTimeMS = results.reduce(function(sum, result) {
+    return result.processingTimeMS === undefined
```
samouss: For which reason the `processingTimeMS` will not be there? I think it's always there (see [doc](https://www.algolia.com/doc/api-reference/api-methods/search/#method-response-processingtimems)).

test/spec/SearchResults/initialization.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+var SearchParameters = require('../../../src/SearchParameters');
+var SearchResults = require('../../../src/SearchResults');
+
+test('processingTime should be the sum of all individual times', function() {
```
samouss: Is it sure that those queries are perform in sequence and not in parallel?

src/SearchResults/index.js
Haroenv: @tkrugg pointed out you can filter it out: https://www.algolia.com/doc/api-reference/api-parameters/responseFields/?language=js#usage-notes

test/spec/SearchResults/initialization.js
Haroenv: I'm not sure what you mean

samouss: It's not related to your PR but more on the global idea to sum up the time. It makes sense to do a sum only if the queries are triggered in sequence. But are we sure of that? Worth asking we won't change the implementation but would be interesting to know.

Haroenv: Good point, we can ask about that

PR #689, opened 2019-05-06
refactor(lodash): compact

Haroenv:
I added tests for this function, but none of the existing real tests test that compact is needed. I'm not too sure if it really is needed in the first place.

I left it for backwards-compatibility, but all tests pass with identity instead of compact

ref for when this was added (without tests 🙌 https://github.com/algolia/algoliasearch-helper-js/pull/45/commits/744da6d1fe1c627ede06c6cfefa8d65c5a9e82b6)

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -1,7 +1,5 @@
 'use strict';
 
-var compact = require('lodash/compact');
-
 var orderBy = require('lodash/orderBy');
 
 var defaults = require('lodash/defaults');
@@ -14,6 +12,7 @@ var partialRight = require('lodash/partialRight');
 
 var find = require('../functions/find');
 var findIndex = require('../functions/findIndex');
+var compact = require('../functions/compact');
 var formatSort = require('../functions/formatSort');
 
 var generateHierarchicalTree = require('./generate-hierarchical-tree');
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -10,6 +10,7 @@ var isFunction = require('lodash/isFunction');
 var partial = require('lodash/partial');
 var partialRight = require('lodash/partialRight');
 
+var compact = require('../functions/compact');
 var find = require('../functions/find');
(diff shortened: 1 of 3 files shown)


src/functions/compact.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = function compact(array) {
+  if (!Array.isArray(array)) {
+    return [];
+  }
+
+  return array.filter(function(element) {
```
eunjae-lee: What about this?

```suggestion
  return array.filter(Boolean);
```

Haroenv: I wasn't too sure about the compatibility of using Boolean without new

eunjae-lee: You were already using `Boolean` without new :)

```return Boolean(element);```

Anyway, I guess it's okay to do so: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean

File: test/spec/functions/compact.js [hunk 2]
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+var compact = require('../../../src/functions/compact');
+
+test('compact removes falsy values from arrays', function() {
+  expect(compact([2])).toEqual([2]);
+  expect(compact([2, false, null, undefined])).toEqual([2]);
+  expect(compact([2, '45', 44])).toEqual([2, '45', 44]);
```
francoischalifour: `0` and `''` are missing from the test (see [Lodash test usage](https://lodash.com/docs/4.17.11#compact)).

src/SearchResults/index.js
```diff
@@ -12,7 +10,7 @@ var isFunction = require('lodash/isFunction');
 var partial = require('lodash/partial');
 var partialRight = require('lodash/partialRight');
 
-var find = require('../functions/find');
+var find = require('../functions/find');
```
Haroenv:
```suggestion
var find = require('../functions/find');
```

PR #690, opened 2019-05-06
refactor(lodash): remove startsWith

Haroenv:
The comparison is a single letter, and the first part is also a string (because it's the value from find), meaning that they can simply be compared I think

### DIFF
--- a/src/functions/formatSort.js
+++ b/src/functions/formatSort.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var find = require('lodash/find');
-var startsWith = require('lodash/startsWith');
 
 /**
  * Transform sort format from user friendly notation to lodash format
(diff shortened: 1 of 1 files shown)


src/functions/formatSort.js
```diff
@@ -14,7 +13,7 @@ module.exports = function formatSort(sortBy, defaults) {
       var sortInstructions = sortInstruction.split(':');
       if (defaults && sortInstructions.length === 1) {
         var similarDefault = find(defaults, function(predicate) {
-          return startsWith(predicate, sortInstruction[0]);
+          return predicate[0] === sortInstruction[0];
```
samouss: On a single letter? Really? You don't think it's a mistake? 

It would have made more sense to use `sortInstructions` with an `S` rather than `sortInstruction`. WDYT?

Haroenv: I was doubting about this too, but forgot when merging, let's rethink this

PR #692, opened 2019-05-07
refactor(lodash): replace "defaults" with pure alternative

Haroenv:
The original lodash tests don't check that it also modifies the target, and as far as I can tell, we don't rely on that behaviour.

### DIFF
--- a/src/functions/defaultsPure.js
+++ b/src/functions/defaultsPure.js
@@ -1,0 +1,15 @@
+'use strict';
+
+
+// NOTE: this behaves like lodash/defaults, but doesn't mutate the target
+module.exports = function defaultsPure() {
+  const sources = Array.prototype.slice.call(arguments);
+  return sources.reduceRight(function(acc, source) {
+    Object.keys(Object(source)).forEach(function(key) {
+      if (source[key] !== undefined) {
+        acc[key] = source[key];
+      }
+    });
+    return acc;
+  }, {});
+};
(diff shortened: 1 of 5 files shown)


test/spec/functions/defaultsPure.js
```diff
@@ -0,0 +1,66 @@
...
+it('should overwrite `undefined` values', function() {
+  var actual = defaults({'a': undefined}, {'a': 1});
+  expect(actual.a).toBe(1);
+});
+
+it('should assign `undefined` values', function() {
+  var source = {'a': undefined, 'b': 1};
+  var actual = defaults({}, source);
+
+  expect(actual).toEqual({'a': undefined, 'b': 1});
```
eunjae-lee: Shouldn't this be like this?
```js
expect(actual).toEqual({'b': 1});
```

Haroenv: this test was copied from lodash, but explicitly putting undefined on an object still matches toEqual I think

eunjae-lee: Ah okay. I just tested and this passes.

```js
expect({ b: 1 }).toEqual({ a: undefined, b: 1 });
```

src/functions/defaultsPure.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+
```
Haroenv:
```suggestion
```

PR #694, opened 2019-05-07
refactor(lodash): merge

Haroenv:
review notes:

- [x] agree on the skipped tests, and remove them
- [x] agree on the notes with the tests, and remove them
- [x] agree on the implementation


src/functions/merge.js
```diff
@@ -0,0 +1,39 @@
...
+      return acc;
+    }
+
+    if (source === undefined) {
+      return acc;
+    }
+
+    if (acc === undefined) {
+      // this causes the source to be augmented, not allowed!
+      return source;
```
Haroenv: I'm not sure if I should make a copy of source or something?

test/spec/functions/mergePure.js
```diff
@@ -0,0 +1,240 @@
...
+  expect(actual).toEqual(Foo);
+  expect(Foo.a).toEqual(1);
+});
+
+it('should merge first source object properties to function', function() {
+  var fn = function() {};
+  var object = {prop: {}};
+  var actual = merge({prop: fn}, object);
+
+  expect(actual.prop).toBeInstanceOf(Function);
```
eunjae-lee: Can you explain to me why actual.prop should be a function instead of {} ?

Haroenv: if the source is a function, we want it to stay as a function. Not too important for our use case, but I didn't beak it

eunjae-lee: I see. then why does [this](https://github.com/algolia/algoliasearch-helper-js/pull/694/commits/4f1ab4d5e4a23b7489525706332c384b794e5c25#diff-e96e00593d1b2b63d3b8689232becc30R77) replace a function with {} ? It's a bit confusing to me.

Haroenv: that was the original test from lodash, it also didn't make sense to me, so I changed it: https://github.com/algolia/algoliasearch-helper-js/pull/694/files#diff-9de5a55358378fba3c54e69c20f364f9R79

eunjae-lee: thanks for the clarification. I was reading from "commits" tab and apparently it wasn't showing me the latest force-pushed commits.

src/requestBuilder.js
```diff
@@ -72,7 +72,7 @@ var requestBuilder = {
       additionalParams.numericFilters = numericFilters;
     }
 
-    return merge(state.getQueryParams(), additionalParams);
+    return merge({}, state.getQueryParams(), additionalParams);
```
eunjae-lee: What happens without prepending {} ?

Haroenv: This was when I initially wasn't sure whether to modify the target object. None of the usages rely on this though, so I made that clear by creating a fresh object. (like you'd do in Object.assign usually)

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+  var actual = merge(Foo, source);
+
+  expect(actual).toEqual(Foo);
+  expect(Foo.a).toEqual(1);
+});
+
+it('should merge first source object properties to function', function() {
+  var fn = function() {};
+  var object = {prop: {}};
+  var actual = merge({prop: fn}, object);
```
samouss: Is it the expected behaviour? I would have thought that the object replaces the function.

```js
_.merge({ props() {} }, { props: {} }) // { props: {} } 
```

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+
+  expect(actual.prop).toBeInstanceOf(Function);
+});
+
+// TODO: differs from lodash, but seems to make more sense to me
+it('should merge first and second source object properties to function', function() {
+  var fn = function() {};
+  var object = {prop: {dogs: 'out'}};
+  var actual = merge({prop: fn}, {prop: fn}, object);
```
samouss: SEE

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+
+  expect(actual.prop).toBe(fn);
+  expect(actual.prop.dogs).toBe('out');
+});
+
+// TODO: if a source was a function, it _will_ get added properties (no big deal for us)
+it.skip('should not merge onto function values of sources', function() {
```
samouss: Since we skip it we can remove it.

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+    characters: [
+      {name: 'barney', age: 36, height: '5\'4"'},
+      {name: 'fred', age: 40, height: '5\'5"'}
+    ]
+  };
+
+  expect(merge(names, ages, heights)).toEqual(expected);
+});
+
+it.skip('should merge sources containing circular references', function() {
```
samouss: Since we skip it we can remove it.

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+
+  var object = new Foo();
+  var actual = merge(object, {a: 1});
+
+  expect(actual).toBe(object);
+  expect(object.a).toBe(1);
+});
+
+// TODO: different from lodash, but seems not necessary
+it.skip('should treat sparse array sources as dense', function() {
```
samouss: Since we skip it we can remove it.

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+  expect('1' in actual).toBe(true);
+  expect(actual).toEqual(expected);
+});
+
+it('should assign `null` values', function() {
+  var actual = merge({a: 1}, {a: null});
+  expect(actual.a).toBe(null);
+});
+
+it.skip('should assign non array/object source values directly', function() {
```
samouss: Since we skip it we can remove it.

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+
+  var actual = values.map(function(value) {
+    var object = merge({}, {a: value, b: {c: value}});
+    return object.a === value && object.b.c === value;
+  });
+
+  expect(actual).toEqual(expected);
+});
+
+it('should not augment source objects 1', function() {
```
samouss: We can find better name than 1.

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+  var source1 = {a: [{a: 1}]};
+  var source2 = {a: [{b: 2}]};
+  var actual = merge({}, source1, source2);
+
+  expect(source1.a).toEqual([{a: 1}]);
+  expect(source2.a).toEqual([{b: 2}]);
+  expect(actual.a).toEqual([{a: 1, b: 2}]);
+});
+
+it('should not augment source objects 2', function() {
```
samouss: We can find a better name than 2.

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+  var object = {a: {'1': 'y', 'b': 'z', 'length': 2}};
+  var actual = merge(object, {a: ['x']});
+
+  expect(actual).toEqual({a: ['x', 'y']});
+
+  actual = merge({a: {}}, {a: []});
+  expect(actual).toEqual({a: []});
+});
+
+it('should not convert strings to arrays when merging arrays of `source`', function() {
```
samouss: We probably don’t have to test this case it’s obvious that we don’t support this kind of operation.

test/spec/functions/merge.js
```diff
@@ -0,0 +1,244 @@
...
+      pass = false;
+    }
+  });
+
+  merge(object, object);
+  expect(pass).toBe(true);
+});
+
+// NOT: we don't want this behavior
+it.skip('should convert values to arrays when merging arrays of `source`', function() {
```
samouss: We can remove it since we skip it. A better alternative would be to write a test that confirm the negation of the current test.

test/spec/functions/merge.js
Haroenv: not if the target is a function apparently, this test is the same as in lodash

samouss: _.merge({ props() {} }, { props: {} })

test/spec/functions/merge.js
Haroenv: likely still  useful, since ES2015 returns keys of strings as if they are arrays, and we definitely don't want to merge strings (it will throw errors)

test/spec/functions/merge.js
Haroenv: b8ce460

test/spec/functions/merge.js
Haroenv: b8ce460

test/spec/functions/merge.js
Haroenv: b8ce460

test/spec/functions/merge.js
Haroenv: b8ce460

test/spec/functions/merge.js
Haroenv: b8ce460

test/spec/functions/merge.js
Haroenv: 73ea23e

test/spec/functions/merge.js
Haroenv: 73ea23e

test/spec/functions/merge.js
Haroenv: lodash behaves differently if a function is top-level or as a value of the object, we don't. Since we don't have functions in the first place, this isn't breaking any usages.

PR #696, opened 2019-05-07
refactor(lodash): intersection

Haroenv:
These are both containing values which are able to be compared by value, so this should be equivalent.

It's already covered under existing tests, because both failed when implementing it wrong on purpose

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var keys = require('lodash/keys');
-var intersection = require('lodash/intersection');
 var forOwn = require('lodash/forOwn');
 var isNaN = require('lodash/isNaN');
 var isEmpty = require('lodash/isEmpty');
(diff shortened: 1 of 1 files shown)


src/SearchParameters/index.js
```diff
@@ -1217,9 +1216,11 @@ SearchParameters.prototype = {
    */
   getRefinedDisjunctiveFacets: function getRefinedDisjunctiveFacets() {
     // attributes used for numeric filter can also be disjunctive
-    var disjunctiveNumericRefinedFacets = intersection(
-      keys(this.numericRefinements),
-      this.disjunctiveFacets
+    var numericRefinements = Object.keys(this.numericRefinements);
+    var disjunctiveNumericRefinedFacets = this.disjunctiveFacets.filter(
```
tkrugg: I think I would have factorised this intersection logic in a function an call it `intersection`, as it's much clearer to understand what's happening like that I believe.

Haroenv: I disagree, I find both usages without the helper function much easier. You might be biased because that's the way it was written before, or because you're very familiar with lodash.

I'm also trying to avoid new helper functions here, because they will quickly become as complicated as the lodash ones, and then we won nothing by removing it :)

src/SearchParameters/index.js
```diff
@@ -1217,9 +1216,11 @@ SearchParameters.prototype = {
...
+      function(facet) {
+        return numericRefinements.indexOf(facet) > -1;
+      }
     );
 
     return keys(this.disjunctiveFacetsRefinements)
```
tkrugg: maybe this is done in another PR but just in case it's not.
` keys(this.disjunctiveFacetsRefinements)` can use `Object.keys` too (just like  line 1219)
```
 var numericRefinements = Object.keys(this.numericRefinements);
```

src/SearchParameters/index.js
tkrugg: I wouldn't say `nothing`. 
The code you would have created is much more concise than the [original] (https://github.com/lodash/lodash/blob/master/.internal/baseIntersection.js), leave alone the dependencies it's using 

It's fine to leave it like that

src/SearchParameters/index.js
Haroenv: #656

PR #698, opened 2019-05-08
refactor(lodash): replace orderBy

Haroenv:
I generated this code:

1. take original lodash code
2. inline imports
3. inline helpers
4. remove edge-case helpers (string instead of array arguments, path instead of element, Symbol support, NaN support etc.)
5. copy tests from lodash
6. change tests to work in Jest
7. remove one test that no longer would have passed

This probably can be simplified a bit more here and there, but likely isn't that important

### DIFF
--- a/src/functions/orderBy.js
+++ b/src/functions/orderBy.js
@@ -1,0 +1,79 @@
+'use strict';
+
+function compareAscending(value, other) {
+  if (value !== other) {
+    var valIsDefined = value !== undefined;
+    var valIsNull = value === null;
+
+    var othIsDefined = other !== undefined;
+    var othIsNull = other === null;
+
+    if (
+      (!othIsNull && value > other) ||
+      (valIsNull && othIsDefined) ||
+      !valIsDefined
+    ) {
+      return 1;
+    }
+    if (
+      (!valIsNull && value < other) ||
+      (othIsNull && valIsDefined) ||
+      !othIsDefined
+    ) {
+      return -1;
+    }
+  }
+  return 0;
+}
+
+/**
+ * @param {Array<object>} collection object with keys in attributes
+ * @param {Array<string>} iteratees attributes
+ * @param {Array<string>} orders asc | desc
+ */
+function orderBy(collection, iteratees, orders) {
+  if (!Array.isArray(collection)) {
+    return [];
+  }
+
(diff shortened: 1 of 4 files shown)


src/functions/orderBy.js
```diff
@@ -0,0 +1,79 @@
...
+      value: value
+    };
+  });
+
+  result.sort(function comparer(object, other) {
+    var index = -1;
+
+    while (++index < object.criteria.length) {
+      var res = compareAscending(object.criteria[index], other.criteria[index]);
+      if (res) {
```
eunjae-lee: We could change this block to 

```js
if (res) {
  if (orders[index] === 'desc') {
    return -res;
  } else {
    return res;
  }
}
```

If `index` is out of range, it will fail `=== 'desc'` anyway.

Haroenv: I was thinking about that, but I think this was added in lodash to shortcut the loop, although reading it again, it might not make a difference

eunjae-lee: hmmm no idea 🤷🏼‍♂️

src/functions/orderBy.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * @param {Array<object>} collection object with keys in attributes
+ * @param {Array<string>} iteratees attributes
+ * @param {Array<string>} orders asc | desc
+ */
+function orderBy(collection, iteratees, orders) {
+  if (!Array.isArray(collection)) {
+    return [];
```
eunjae-lee: We can add a test case for this.

src/functions/orderBy.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  if (!Array.isArray(orders)) {
+    orders = [];
```
eunjae-lee: We can add a test case for this.

Haroenv: this has a test case: https://github.com/algolia/algoliasearch-helper-js/pull/698/files#diff-67016e5575af72c2e6fbd33e554c7988R31

src/functions/orderBy.js
Haroenv: cc320ed3de6798a58e82a151e33f5092cd486112

PR #702, opened 2019-05-09
fix(sortBy): compare whole prefix instead of first character

Haroenv:
This was noticed in https://github.com/algolia/algoliasearch-helper-js/pull/690#discussion_r282467917 by @samouss, so I took the time to rewrite this function a little bit to be more clear (IMO)

### DIFF
--- a/src/functions/formatSort.js
+++ b/src/functions/formatSort.js
@@ -5,22 +5,29 @@ var find = require('./find');
 /**
  * Transform sort format from user friendly notation to lodash format
  * @param {string[]} sortBy array of predicate of the form "attribute:order"
+ * @param {string[]} [defaults] array of predicate of the form "attribute:order"
  * @return {array.<string[]>} array containing 2 elements : attributes, orders
  */
 module.exports = function formatSort(sortBy, defaults) {
+  defaults = (defaults || []).map(function(sort) {
+    return sort.split(':');
+  });
+
   return sortBy.reduce(
-    function preparePredicate(out, sortInstruction) {
-      var sortInstructions = sortInstruction.split(':');
-      if (defaults && sortInstructions.length === 1) {
-        var similarDefault = find(defaults, function(predicate) {
-          return predicate[0] === sortInstruction[0];
+    function preparePredicate(out, sort) {
+      var sortInstruction = sort.split(':');
+      if (sortInstruction.length === 1) {
+        var matchingDefault = find(defaults, function(defaultInstruction) {
+          return defaultInstruction[0] === sortInstruction[0];
         });
-        if (similarDefault) {
-          sortInstructions = similarDefault.split(':');
+        if (matchingDefault) {
+          out[0].push(matchingDefault[0]);
+          out[1].push(matchingDefault[1]);
+          return out;
         }
       }
-      out[0].push(sortInstructions[0]);
-      out[1].push(sortInstructions[1]);
+      out[0].push(sortInstruction[0]);
+      out[1].push(sortInstruction[1]);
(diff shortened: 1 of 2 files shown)


src/functions/formatSort.js
```diff
@@ -5,22 +5,29 @@ var find = require('./find');
 /**
  * Transform sort format from user friendly notation to lodash format
  * @param {string[]} sortBy array of predicate of the form "attribute:order"
+ * @param {string[]} [defaults] array of predicate of the form "attribute:order"
  * @return {array.<string[]>} array containing 2 elements : attributes, orders
  */
 module.exports = function formatSort(sortBy, defaults) {
+  defaults = (defaults || []).map(function(sort) {
```
samouss: Why do we override the value, we can create a new one.

samouss: Not sure the default value is useful, IIRC it's always called with a value.

src/functions/formatSort.js
```diff
@@ -5,22 +5,29 @@ var find = require('./find');
...
+
   return sortBy.reduce(
-    function preparePredicate(out, sortInstruction) {
-      var sortInstructions = sortInstruction.split(':');
-      if (defaults && sortInstructions.length === 1) {
-        var similarDefault = find(defaults, function(predicate) {
-          return predicate[0] === sortInstruction[0];
+    function preparePredicate(out, sort) {
+      var sortInstruction = sort.split(':');
+      if (sortInstruction.length === 1) {
```
samouss: It would ease the reading with the condition inverted. We can early return at the beginning of the function, it would then reduce the nesting of the function.

src/functions/formatSort.js
Haroenv: Hierarchical calls it without defaults I think

src/functions/formatSort.js
```diff
@@ -5,22 +5,38 @@ var find = require('./find');
...
+    function preparePredicate(out, sort) {
+      var sortInstruction = sort.split(':');
+
+      if (sortInstruction.length > 1) {
+        out[0].push(sortInstruction[0]);
+        out[1].push(sortInstruction[1]);
+        return out;
+      }
+
+      var matchingDefault = find(defaultInstructions, function(
```
samouss: We can move this assignation above the `if`. We can then group the condition to only use two `return`. We might compute the matching default for nothing but I don't think it matters that much. The overall function would be easier to understand.

Haroenv: like this?

```js
      var sortInstruction = sort.split(':');
      
      var matchingDefault = find(defaultInstructions, function(
        defaultInstruction
      ) {
        return defaultInstruction[0] === sortInstruction[0];
      });

      if (sortInstruction.length > 1 || !matchingDefault) {
        out[0].push(sortInstruction[0]);
        out[1].push(sortInstruction[1]);
        return out;
      }

      out[0].push(matchingDefault[0]);
      out[1].push(matchingDefault[1]);
      return out;
```

src/functions/formatSort.js
```diff
@@ -5,22 +5,32 @@ var find = require('./find');
...
-      if (defaults && sortInstructions.length === 1) {
-        var similarDefault = find(defaults, function(predicate) {
-          return predicate[0] === sortInstruction[0];
-        });
-        if (similarDefault) {
-          sortInstructions = similarDefault.split(':');
-        }
+    function preparePredicate(out, sort) {
+      var sortInstruction = sort.split(':');
+
```
Haroenv: according to circle this line has trailing whitespace?

PR #703, opened 2019-05-10
feat(SearchParameters): avoid undefined values

samouss:
This PR updates the internals of the `SearchParameters` to avoid `undefined` values inside the search parameters. This change allows to easily merge multiple `SearchParameters` top-down since we don't have "ghost" value on the object. The outcome is the change of semantics for the parameters and setters:

- `query is omit`: params does not have query e.g. `searchBox` not mounted
- `query: ''`: params does have an empty query e.g. `searchBox` mounted with initial state
- `query: 'Apple'`: params does have a query refinement e.g. `searchBox` mounted with refinement

-----

- `setQuery()`: omit the query e.g. `widget.dispose`
- `setQuery('')`: sets the default value for the query e.g. `widget.getConfiguration`
- `setQuery('Apple')`: sets the given value for the query e.g. `widget.refine`

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -197,7 +197,10 @@ function SearchParameters(newParameters) {
 
   var self = this;
   Object.keys(params).forEach(function(paramName) {
-    if (SearchParameters.PARAMETERS.indexOf(paramName) === -1) {
+    var isKeyUnknown = SearchParameters.PARAMETERS.indexOf(paramName) === -1;
+    var isValueDefined = params[paramName] !== undefined;
+
+    if (isKeyUnknown && isValueDefined) {
       self[paramName] = params[paramName];
     }
   });
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -1358,15 +1358,32 @@ SearchParameters.prototype = {
       throw error;
     }
 
-    var parsedParams = SearchParameters._parseNumbers(params);
+    var self = this;
+    var nextWithNumbers = SearchParameters._parseNumbers(params);
+    var previousPlainObject = Object.keys(this).reduce(function(acc, key) {
+      acc[key] = self[key];
+      return acc;
+    }, {});
 
-    return this.mutateMe(function mergeWith(newInstance) {
-      Object.keys(params).forEach(function(k) {
-        newInstance[k] = parsedParams[k];
-      });
+    var nextPlainObject = Object.keys(nextWithNumbers).reduce(
+      function(previous, key) {
+        var isPreviousValueDefined = previous[key] !== undefined;
+        var isNextValueDefined = nextWithNumbers[key] !== undefined;
 
-      return newInstance;
-    });
+        if (isPreviousValueDefined && !isNextValueDefined) {
+          return omit(previous, [key]);
+        }
+
+        if (isNextValueDefined) {
+          previous[key] = nextWithNumbers[key];
+        }
+
+        return previous;
+      },
+      previousPlainObject
+    );
+
+    return new this.constructor(nextPlainObject);
   },
 
   /**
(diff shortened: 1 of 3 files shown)


src/SearchParameters/index.js
```diff
@@ -1327,17 +1330,32 @@ SearchParameters.prototype = {
...
+    var nextWithNumbers = SearchParameters._parseNumbers(params);
+    var previousPlainObject = Object.keys(this).reduce(function(acc, key) {
+      acc[key] = self[key];
+      return acc;
+    }, {});
+
+    var nextPlainObject = Object.keys(nextWithNumbers).reduce(
+      function(previous, key) {
+        var isPreviousValueDefined = previous[key] !== undefined;
+        var isNextValueDefined = nextWithNumbers[key] !== undefined;
```
Haroenv: what about:

```js
previous[key] = nextWithNumbers[key];

if (previous[key] === undefined) {
  return omit(previous, [key]);
}

return previous;
```

Haroenv: Ah no this is wrong because an undefined value would override

src/SearchParameters/index.js
```diff
@@ -1327,17 +1330,32 @@ SearchParameters.prototype = {
...
+          return omit(previous, [key]);
+        }
+
+        if (isNextValueDefined) {
+          previous[key] = nextWithNumbers[key];
+        }
+
+        return previous;
+      },
+      previousPlainObject
```
Haroenv: I wonder if we could have `this` here, then I don't think we need to call new this.constructor anymore, but rather return `this` & ignore the value returned from the reduce (ie use forEach instead)

samouss: We have to call the constructor, we don't want to mutate the `SearchParameters`. We have to return a new instance. The `mutateMe` function despite the name was returning [a new instance](https://github.com/algolia/algoliasearch-helper-js/pull/703/files/4942fbd413afb02cb3b3a17a185c68eb2129422b#diff-ecd72976141723d9448df45e667cb4beL1376) of the `SearchParameters`. With `this` the only solution to "omit" attributes is the usage of the `delete` operator (which might be a bit slower). Without talking about performance I'm more comfortable with this implementation.

src/SearchParameters/index.js
```diff
@@ -197,7 +197,10 @@ function SearchParameters(newParameters) {
 
   var self = this;
   Object.keys(params).forEach(function(paramName) {
-    if (SearchParameters.PARAMETERS.indexOf(paramName) === -1) {
+    var isKeyUnknown = SearchParameters.PARAMETERS.indexOf(paramName) === -1;
```
francoischalifour: Reversing this expression would make it easier to understand.

src/SearchParameters/index.js
```diff
@@ -1393,20 +1413,6 @@ SearchParameters.prototype = {
...
   },
-  /**
-   * Helper function to make it easier to build new instances from a mutating
-   * function
-   * @private
-   * @param {function} fn newMutableState -> previousState -> () function that will
-   * change the value of the newMutable to the desired state
-   * @return {SearchParameters} a new instance with the specified modifications applied
-   */
-  mutateMe: function mutateMe(fn) {
```
francoischalifour: For reference, this is only used in [`connectConfigure` in InstantSearch.js](https://github.com/algolia/instantsearch.js/blob/49a5b64e42deec3453669d7dce788395cf898ba8/src/connectors/configure/connectConfigure.js#L96-L99). One solution for this removal could be to set the `widgetParams` search parameters to `undefined`.

src/SearchParameters/index.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/703/commits/6e705246a09365d3dccd480bc5773f31e24ffb0a

PR #704, opened 2019-05-13
feat(search): allow the search only with Derived Helpers

samouss:
This PR introduces a new method called `searchOnlyWithDerivedHelpers`. As the name suggests the function allows to bypass the "main" instance search to only use the Derived Helpers. The main idea behind the API is to be able to control which state is used by the query, without having to alter the internal state of the helper (it would be useful for federated search for example). With the `.derive` API we have a full control on the query with the derivation function. We can alter the state without impact on the internal instance of the Helper.

I choose to implement this behavior with a dedicated method to limit the number of changes. We'll release a major version but we should keep the support of the original behavior because it will not only benefit InstantSearch.js. We also want to benefit the Lodash reduction in React InstantSearch.

```js
const helper = algoliasearchHelper(client, 'index');
const derivedHelper = helper.derive(state => {
  // Alter the state without impact on the main instance
  return state;
});

helper.on('search', () => {
  // The search is not triggered on the main instance
});

derivedHelper.on('search', () => {
  // The search is only triggered on the derived instance
});

// Trigger only 1 queries vs 2 queries with a regular search
helper.searchOnlyWithDerivedHelpers();
```

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -144,7 +144,12 @@ inherits(AlgoliaSearchHelper, events.EventEmitter);
  * @chainable
  */
 AlgoliaSearchHelper.prototype.search = function() {
-  this._search();
+  this._search({onlyWithDerivedHelpers: false});
+  return this;
+};
+
+AlgoliaSearchHelper.prototype.searchOnlyWithDerivedHelpers = function() {
+  this._search({onlyWithDerivedHelpers: true});
   return this;
 };
 
@@ -1224,28 +1229,33 @@ AlgoliaSearchHelper.prototype.getHierarchicalFacetBreadcrumb = function(facetNam
  * @fires result
  * @fires error
  */
-AlgoliaSearchHelper.prototype._search = function() {
+AlgoliaSearchHelper.prototype._search = function(options) {
   var state = this.state;
-  var mainQueries = requestBuilder._getQueries(state.index, state);
+  var states = [];
+  var mainQueries = [];
 
-  var states = [{
-    state: state,
-    queriesCount: mainQueries.length,
-    helper: this
-  }];
+  if (!options.onlyWithDerivedHelpers) {
+    mainQueries = requestBuilder._getQueries(state.index, state);
 
-  this.emit('search', {
-    state: state,
-    results: this.lastResults
-  });
+    states.push({
+      state: state,
+      queriesCount: mainQueries.length,
+      helper: this
+    });
+
+    this.emit('search', {
+      state: state,
+      results: this.lastResults
+    });
+  }
 
   var derivedQueries = this.derivedHelpers.map(function(derivedHelper) {
     var derivedState = derivedHelper.getModifiedState(state);
-    var queries = requestBuilder._getQueries(derivedState.index, derivedState);
+    var derivedStateQueries = requestBuilder._getQueries(derivedState.index, derivedState);
 
     states.push({
       state: derivedState,
-      queriesCount: queries.length,
+      queriesCount: derivedStateQueries.length,
       helper: derivedHelper
     });
 
@@ -1254,7 +1264,7 @@ AlgoliaSearchHelper.prototype._search = function() {
       results: derivedHelper.lastResults
     });
 
-    return queries;
+    return derivedStateQueries;
   });
 
   var queries = Array.prototype.concat.apply(mainQueries, derivedQueries);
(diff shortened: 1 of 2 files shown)


test/spec/algoliasearch.helper/derive/multiqueries.js
```diff
@@ -2,48 +2,54 @@
...
+}
+
+test('[Derived helper] no derived helpers', function() {
+  var client = makeFakeClient(assertions);
   var helper = algoliasearchHelper(client, '');
+
   helper.search();
 
-  function searchTest(requests) {
+  function assertions(requests) {
```
Haroenv: I think I'd find these easier to read if they were inlined in the creation of the fake client instead of being hoisted, wdyt?

src/algoliasearch.helper.js
```diff
@@ -144,7 +144,12 @@ inherits(AlgoliaSearchHelper, events.EventEmitter);
  * @chainable
  */
 AlgoliaSearchHelper.prototype.search = function() {
-  this._search();
+  this._search({onlyWithDerivedHelpers: false});
+  return this;
+};
+
+AlgoliaSearchHelper.prototype.searchOnlyWithDerivedHelpers = function() {
```
Haroenv: I was thinking about not exposing this one externally, and relying on `._search({onlyWithDerivedHelpers: false})` in InstantSearch, but that probably doesn't really bring much value

test/spec/algoliasearch.helper/derive/multiqueries.js
samouss: I've kept the original implementation. At first, I've changed it but I revert it. I've found it easier like this. Most of the time we expect to have the assertions at the end of the test. I don't have a strong opinion about that though. We can change it.

src/algoliasearch.helper.js
samouss: We can avoid exposing a function but we can't use your solution. For the federated search implementation, we have to override the `.search` method of the children Helper to bubbles up until the root. With the private call, we can't easily intercept the call anymore. What we could do is to expose an option rather than a dedicated method:

```js
AlgoliaSearchHelper.prototype.search = function(options) {
  this._search(options);
  return this;
};
```

I don't think, we'll document the method anyway. This solution is short term. Once we have the federated search implementation, it would be a good opportunity to redesign a "Helper" with a first-class support of it. At the end both implementations are equivalent. The first one is a bit more explicit IMO.

test/spec/algoliasearch.helper/derive/multiqueries.js
```diff
@@ -2,48 +2,54 @@
...
     expect(requests.length).toBe(1);
-
-    return new Promise(function() {});
   }
 });
 
 test('[Derived helper] 1 derived helpers, no modifications', function() {
```
francoischalifour:
```suggestion
test('[Derived helper] 1 derived helper, no modifications', function() {
```

test/spec/algoliasearch.helper/derive/multiqueries.js
```diff
@@ -2,48 +2,54 @@
...
-    return new Promise(function() {});
   }
 });
 
-test('[Derived helper] no derived helpers, modification', function() {
-  expect.assertions(4);
-  var client = {
-    search: searchTest
-  };
+test('[Derived helper] 1 derived helpers, modification', function() {
```
francoischalifour:
```suggestion
test('[Derived helper] 1 derived helper, modification', function() {
```

test/spec/algoliasearch.helper/derive/multiqueries.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/704/commits/ed5d46b16692453c068c6f1385f48e3e552b2362

test/spec/algoliasearch.helper/derive/multiqueries.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/704/commits/ed5d46b16692453c068c6f1385f48e3e552b2362

test/spec/algoliasearch.helper/derive/multiqueries.js
```diff
@@ -2,48 +2,54 @@
...
+  return {
+    search: function() {
+      assertions.apply(null, arguments);
+
+      return new Promise(function() {});
+    }
   };
+}
+
+test('trigger a search without without derivation', function() {
```
francoischalifour:
```suggestion
test('trigger a search without derivation', function() {
```

PR #707, opened 2019-05-14
feat(getState): remove "filter" option

### LINKED ISSUE #534: how do I know what filters I can use in getState()
I was viewing the docs for `getState();`

and it said this about the filters: _optional filters to retrieve only a subset of the state_

So where can I find out what filters I can use and how I should spell the name?

I needed to print out my text query together with my numeric refinements and my current page in the pagination and index name. I wasnt sure if the pagionation was named p or page but I tried this which seems to work:

`getState(['query', 'attribute:*', 'page', 'index']));`

So where can I find what filters I can use and how I should spell their name?

Haroenv:
This also removes the `.filter` function on `SearchParameters` which isn't used.

Another slight reduction in size: 11.2  -> 10.9 (after lodash.merge)

closes #534

BREAKING CHANGE: getState(filters) is replaced my manually filtering the returned object
BREAKING CHANGE: SearchParameters.filter is removed


Haroenv:
looks like I'll need to remove a part of the docs

PR #709, opened 2019-05-15
fix(merge): consider "null" as not an object

Haroenv:
This behaves the same as lodash (but they didn't have a test for it)

This fixes the behaviour seen in http://localhost:6006/?path=/story/breadcrumb--with-hierarchical-menu on `refactor/usage-search-parameters` (on InstantSearch.js when you  link this version of the helper)

IFW-777

### DIFF
--- a/src/functions/merge.js
+++ b/src/functions/merge.js
@@ -1,11 +1,11 @@
 'use strict';
 
-function isPlainObject(value) {
+function isObjectLike(value) {
   return typeof value === 'object' && value !== null;
 }
 
 function clone(value) {
-  if (typeof value === 'object' && value !== null) {
+  if (isObjectLike(value)) {
     return merge(Array.isArray(value) ? [] : {}, value);
   }
   return value;
@@ -43,11 +43,11 @@ function merge() {
       return clone(source);
     }
 
-    if (!isPlainObject(acc) && typeof acc !== 'function') {
+    if (!isObjectLike(acc) && typeof acc !== 'function') {
       return clone(source);
     }
 
-    if (!isPlainObject(source) && typeof source !== 'function') {
+    if (!isObjectLike(source) && typeof source !== 'function') {
       return acc;
     }
 
--- a/src/functions/merge.js
+++ b/src/functions/merge.js
@@ -1,5 +1,9 @@
 'use strict';
 
+function isPlainObject(value) {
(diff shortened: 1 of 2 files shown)


test/spec/functions/merge.js
```diff
@@ -171,3 +171,9 @@ it('should not convert strings to arrays when merging arrays of `source`', funct
   expect(actual).toEqual({a: ['x', 'y', 'z']});
 });
 
+it('should return an object for null', function() {
```
Haroenv:
```suggestion
it('should not consider null as an object', function() {
```

test/spec/functions/merge.js
```diff
@@ -171,3 +171,9 @@ it('should not convert strings to arrays when merging arrays of `source`', funct
   expect(actual).toEqual({a: ['x', 'y', 'z']});
 });
 
+it('should not consider null as an object', function() {
+  expect(merge(null, {})).toEqual({});
+  expect(merge({this: 'b'}, {this: 'k'}, {this: null})).toEqual({this: null});
+  expect(merge({this: 'b'}, {this: null}, {this: undefined})).toEqual({this: null});
+  expect(merge({this: 'b'}, {this: null}, {this: 'k'})).toEqual({this: 'k'});
```
tkrugg: I think it might be relevant to add a test with `new RegExp()` or `new Date()`, to check the this spec trait: `- treats non-plain objects as plain`
If this merge put in a place such as `functions/` it might end up being reused in theses situations.

src/functions/merge.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+function isPlainObject(value) {
```
tkrugg: This test is actually called `isObjectLike` in lodash. I'm think it might not be a bad idea to keep lodash naming when the implem is the same. I'll let you the judge of that.
https://github.com/lodash/lodash/blob/master/isObjectLike.js#L23

src/functions/merge.js
```diff
@@ -1,5 +1,9 @@
...
+  return typeof value === 'object' && value !== null;
+}
+
 function clone(value) {
   if (typeof value === 'object' && value !== null) {
```
tkrugg: did you mean to use isPlainObject here too?

PR #711, opened 2019-05-16
chore: remove Bower support

samouss:


samouss:
@Haroenv for which use cases?

PR #712, opened 2019-05-16
chore(release): allow canary release

samouss:
This PR allows to release a "canary" version of the package. 

The version follows this format: `0.0.0-SHORT_SHA-1` where `SHORT_SHA-1` is the hash of the latest commit. The version is then published on npm with the tag `canary`. The command does not update the CHANGELOG or create a Git Tag. The usage of the version is mostly internal. Its purpose is to publish a non-stable version that can be used directly from other projects via npm (useful for the CI).

At some point, we might want to port the process on other projects as well. It would allow us to test packages directly from E2E tests or even on example projects that we maintain. I didn't use it (yet) to publish a version (I did test it though with the `pack` command). I'll do it once the PR is approved and merged.

Screenshot 2019-05-17 at 14 30 40

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -31,20 +31,20 @@ if (isCanary) {
 }
 
 function releaseCanaryVersion() {
-  shell.echo(`Algoliasearch-Helper release CANARY version`);
+  shell.echo(`Algolia JS Helper release CANARY version`);
 
   checkCleanWorkdir();
   updateCanaryVersion(canaryVersion => {
     build();
     publishOnNpm('canary');
     revertStableVersion(packageJson.version, () => {
-      shell.echo(`Algoliasearch-Helper v${canaryVersion} released`);
+      shell.echo(`Algolia JS Helper v${canaryVersion} released`);
     });
   });
 }
 
 function releaseStableVersion() {
-  shell.echo(`Algoliasearch-Helper release STABLE version`);
+  shell.echo(`Algolia JS Helper release STABLE version`);
 
   checkDevelopBranch();
   checkCleanWorkdir();
@@ -58,7 +58,7 @@ function releaseStableVersion() {
       publish();
       goBackToDevelop();
 
-      shell.echo(`Algoliasearch-Helper v${version} released`);
+      shell.echo(`Algolia JS Helper v${version} released`);
     });
   });
 }
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -1,9 +1,11 @@
 #!/usr/bin/env node
+
 'use strict';
 
 const prompt = require('prompt');
 const semver = require('semver');
 const mversion = require('mversion');
+const argv = require('yargs').argv;
 const path = require('path');
 const fs = require('fs');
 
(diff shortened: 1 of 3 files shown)


scripts/release.js
```diff
@@ -20,29 +22,57 @@ const packageJson = require('../package.json');
...
   });
-});
+}
 
-function checkEnvironment() {
+function releaseStableVersion() {
+  shell.echo(`Algoliasearch-Helper release STABLE version`);
+
+  checkDevelopBranch();
+  checkCleanWorkdir();
```
samouss: Those two functions are the equivalent of the previous call `checkEnvironment()`.

scripts/release.js
```diff
@@ -136,3 +170,16 @@ function goBackToDevelop() {
...
+function updateCanaryVersion(cb) {
+  const lastCommitHash = shell.exec('git rev-parse --short HEAD', {silent: true}).toString().trim();
+  const canaryVersion = `0.0.0-${lastCommitHash}`;
+
+  bumpVersion(canaryVersion, () => {
+    cb(canaryVersion);
+  });
+}
+
+function revertStableVersion(version, cb) {
```
samouss: I've used a function rather than inlining the call to be more explicit.

scripts/release.js
```diff
@@ -20,29 +22,57 @@ const packageJson = require('../package.json');
...
+  releaseStableVersion();
+}
 
-checkEnvironment();
-mergeDevIntoMaster();
-showChangelog(shell);
-promptVersion(packageJson.version, (version) => {
-  bumpVersion(version, () => {
+function releaseCanaryVersion() {
+  shell.echo(`Algoliasearch-Helper release CANARY version`);
```
francoischalifour:
```suggestion
  shell.echo(`AlgoliaSearch Helper release CANARY version`);
```

scripts/release.js
```diff
@@ -20,29 +22,57 @@ const packageJson = require('../package.json');
...
+  checkCleanWorkdir();
+  updateCanaryVersion(canaryVersion => {
     build();
-    updateChangelog(shell);
-    commitNewFiles(version);
-    publish();
-    goBackToDevelop();
+    publishOnNpm('canary');
+    revertStableVersion(packageJson.version, () => {
+      shell.echo(`Algoliasearch-Helper v${canaryVersion} released`);
```
francoischalifour:
```suggestion
      shell.echo(`AlgoliaSearch Helper v${canaryVersion} released`);
```

scripts/release.js
```diff
@@ -20,29 +22,57 @@ const packageJson = require('../package.json');
...
+    });
   });
-});
+}
 
-function checkEnvironment() {
+function releaseStableVersion() {
+  shell.echo(`Algoliasearch-Helper release STABLE version`);
```
francoischalifour:
```suggestion
  shell.echo(`AlgoliaSearch Helper release STABLE version`);
```

scripts/release.js
```diff
@@ -20,29 +22,57 @@ const packageJson = require('../package.json');
...
+  showChangelog(shell);
+  promptVersion(packageJson.version, (version) => {
+    bumpVersion(version, () => {
+      build();
+      updateChangelog(shell);
+      commitNewFiles(version);
+      publish();
+      goBackToDevelop();
+
+      shell.echo(`Algoliasearch-Helper v${version} released`);
```
francoischalifour:
```suggestion
      shell.echo(`AlgoliaSearch Helper v${version} released`);
```

scripts/release.js
```diff
@@ -20,29 +22,57 @@ const packageJson = require('../package.json');
...
+
+  checkCleanWorkdir();
+  updateCanaryVersion(canaryVersion => {
     build();
-    updateChangelog(shell);
-    commitNewFiles(version);
-    publish();
-    goBackToDevelop();
+    publishOnNpm('canary');
+    revertStableVersion(packageJson.version, () => {
```
francoischalifour: Is this function completely resetting the state it was in before creating the canary release? (`version` in `package.json`, git staging, etc.)

It's quite hard to follow the release flow.

samouss: Yes, it does. I agree that it's not the optimal solution. A better approach would be to build the complete package into a dedicated folder that we can discard at the end. Unfortunately, that would completely change the current release process. I tried to minimize the changes, hence the update + revert of the changes.

scripts/release.js
Haroenv: the package name is `algoliasearch-helper` though

scripts/release.js
```diff
@@ -136,3 +170,16 @@ function goBackToDevelop() {
 function build() {
   shell.exec('yarn run build');
 }
+
+function updateCanaryVersion(cb) {
+  const lastCommitHash = shell.exec('git rev-parse --short HEAD', {silent: true}).toString().trim();
+  const canaryVersion = `0.0.0-${lastCommitHash}`;
```
Haroenv: as mentioned, this isn't incrementing, although I'm not too sure if that's an issue

scripts/release.js
Haroenv: here it's still the old version, since you require before you change things right?

scripts/release.js
francoischalifour: We should either go with the package name ("algoliasearch-helper") or the library title ("AlgoliaSearch Helper").

samouss: Yep, will use the name that we have on the Community website which is: Algolia JS Helper.

scripts/release.js
samouss: Yes it is.

scripts/release.js
samouss: I've tested on a package and it seems to work as expected. I manage to publish three versions on the `canary` tag with three different commit messages. I've checked that the hash does not follow each other in any way. It looks like it's working.

> https://yarnpkg.com/en/package/@samouss/useless

scripts/release.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/712/commits/6ebe9c7a44f33183c3cf04afc3285369663cebb6

scripts/release.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/712/commits/6ebe9c7a44f33183c3cf04afc3285369663cebb6

scripts/release.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/712/commits/6ebe9c7a44f33183c3cf04afc3285369663cebb6

scripts/release.js
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/712/commits/6ebe9c7a44f33183c3cf04afc3285369663cebb6

francoischalifour:
Could we rely on a build number that is incremental? (or a hashed date)

The commit hash is great but there's no easy way to know if a build is more recent than another.

For example, [the dumb package released on unpkg](https://unpkg.com/@samouss/useless/) isn't listed based on the date. It's not easy to know which one is older.

Haroenv:
on Yarn the list has a date though, I guess that's enough to work with, since usually you just want to use a specific version I guess?

francoischalifour:
That's right. Having this info on the Yarn website is OK for me!

PR #714, opened 2019-05-20
fix(merge): use `deepmerge` instead

Haroenv:
Quite a lot of the behaviour isn't as it was with lodash.merge or with my earlier merge implementation, but it's at least fixing the bugs we had in InstantSearch AFAICT


Haroenv:
it's quite likely that this isn't really implemented correctly, it's more of a WIP to allow you to try it out in InstantSearch

PR #716, opened 2019-06-18
 fix(merge): change implementation 

tkrugg:
- split the implem into merge(target, ...sources) and _merge(target, source), I think it's more simple this way.
- I reorganized some of the conditions

The description and the test suit should not change. Implementation has changed but behaviour and limitation are the same (Note that the lodash inspired test suite for merge should be passing).

Other minor change to note: use .toStrictEqual instead of .toEqual in tests
closes https://github.com/algolia/algoliasearch-helper-js/pull/714  https://github.com/algolia/algoliasearch-helper-js/pull/709

### DIFF
--- a/src/functions/merge.js
+++ b/src/functions/merge.js
@@ -7,7 +7,7 @@ function clone(value) {
   return value;
 }
 
-function isObjectOrArray(value) {
+function isObjectOrArrayOrFunction(value) {
   return (
     typeof value === 'function' ||
     Array.isArray(value) ||
@@ -32,7 +32,7 @@ function _merge(target, source) {
       continue;
     }
 
-    if (isObjectOrArray(targetVal) && isObjectOrArray(sourceVal)) {
+    if (isObjectOrArrayOrFunction(targetVal) && isObjectOrArrayOrFunction(sourceVal)) {
       target[key] = _merge(targetVal, sourceVal);
     } else {
       target[key] = clone(sourceVal);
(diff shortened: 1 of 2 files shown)


src/functions/merge.js
```diff
@@ -2,11 +2,45 @@
...
+
+  for (var key in source) {
+    if (!Object.prototype.hasOwnProperty.call(source, key)) {
+      continue;
+    }
+
+    var sourceVal = source[key];
+    var targetVal = target[key];
+
+    if (typeof  targetVal !== 'undefined' && typeof sourceVal === 'undefined') {
```
yannickcr:
```suggestion
    if (typeof targetVal !== 'undefined' && typeof sourceVal === 'undefined') {
```

src/functions/merge.js
```diff
@@ -2,11 +2,45 @@
...
   if (typeof value === 'object' && value !== null) {
-    return merge(Array.isArray(value) ? [] : {}, value);
+    return _merge(Array.isArray(value) ? [] : {}, value);
   }
   return value;
 }
 
+function isObjectOrArray(value) {
+  return (
+    typeof value === 'function' ||
```
yannickcr: I think it would be clearer for the reader to reflect in the function's name that we also return `true` for functions.
`isObjectOrArrayOrFunction` ?

tkrugg: done here: 
87c9024

PR #718, opened 2019-06-21
fix(lodash/intersection): replace with custom implementation

tkrugg:
I reverted the commit replacing intersection than copie the current implem into a new `intersection` function
I add something to avoid duplicates and some tests.

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -2,6 +2,7 @@
 
 var merge = require('../functions/merge');
 var defaultsPure = require('../functions/defaultsPure');
+var intersection = require('lodash/intersection');
 var find = require('../functions/find');
 var valToNumber = require('../functions/valToNumber');
 var omit = require('../functions/omit');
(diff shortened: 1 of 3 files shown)


src/SearchParameters/index.js
```diff
@@ -1265,16 +1264,12 @@ SearchParameters.prototype = {
    * @return {string[]}
    */
   getRefinedHierarchicalFacets: function getRefinedHierarchicalFacets() {
-    var hierarchicalRefinedFacets = Object.keys(
-      this.hierarchicalFacetsRefinements
+    return intersection(
+      // enforce the order between the two arrays,
```
samouss: I'm not completely sure to understand the comment. What does it mean?

Haroenv:
I'm sure this has a reason, but what was wrong with the version without helper function?

samouss:
@Haroenv The previous version didn't remove the duplicates, but the Lodash version does.

src/SearchParameters/index.js
Haroenv: This comment was there originally by the way

samouss: Ah. It doesn't make sense to me anymore.

Haroenv:
maybe best to also add a test to its usage to show what this is fixing

src/SearchParameters/index.js
Haroenv: I think we can remove this comment here

tkrugg: The output is an array that contains elements from both passed arrays, so we might wonder which of the two arrays dictates the order.

In the case of [lodash intersection function](https://lodash.com/docs/4.17.11#intersection), it's the first array. 
Example:
```js
intersection([1, 2, 3, 4, 5], [5, 3]) === [3, 5] 
intersection([1, 2, 3, 4, 5], [3, 5]) === [3, 5]
intersection([1, 2, 5, 4, 3], [3, 5]) === [5, 3]
```


Whoever used intersection here wanted to keep the order of `this.hierarchicalFacets.map(...)` over the the order of `Object.keys(this.hierarchicalFacetsRefinements)`.

PR #720, opened 2019-07-10
fix(getFacetValues): don't throw error when there's no facet

Haroenv:
connectMenu could throw if a facet is retrieved which isn't in the SearchParameters, but we are changing that to return undefined in those cases.

This is part 1 of the big error removal process.

see algolia/instantsearch.js#3084
see #722

### DIFF
--- a/test/spec/SearchResults/getFacetValues.js
+++ b/test/spec/SearchResults/getFacetValues.js
@@ -149,7 +149,7 @@ test('getFacetValues(disjunctive) returns correct facet values with the name `le
   expect(facetValues.length).toBe(2);
 });
 
-test.only('getFacetValues(unknown) returns undefined (does not throw)', function() {
+test('getFacetValues(unknown) returns undefined (does not throw)', function() {
   var searchParams = new SearchParameters({
     index: 'instant_search'
   });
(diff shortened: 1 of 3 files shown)


test/spec/SearchResults/getFacetValues.js
```diff
@@ -148,3 +148,20 @@ test('getFacetValues(disjunctive) returns correct facet values with the name `le
   expect(facetValues).toEqual(expected);
   expect(facetValues.length).toBe(2);
 });
+
+test.only('getFacetValues(unknown) returns undefined (does not throw)', function() {
```
samouss:
```suggestion
test('getFacetValues(unknown) returns undefined (does not throw)', function() {
```

test/spec/SearchResults/getFacetValues.js
```diff
@@ -148,3 +148,20 @@ test('getFacetValues(disjunctive) returns correct facet values with the name `le
...
+  var result = {
+    query: '',
+    // it does not matter if the result here is given or not,
+    // if something is not a parameter, it will not be read.
+    facets: {}
+  };
+
+  var results = new SearchResults(searchParams, [result, result]);
+
+  expect(results.getFacetValues('type')).toEqual(undefined);
```
samouss:
```suggestion
  expect(results.getFacetValues('type')).toBeUndefined();
```

Issue #722, opened 2019-07-10
Remove unneeded errors


Haroenv:
We are removing those errors either with undefined if no suitable value gets found, or with the suitable default value for that type. These errors are removed because there's no meaningful difference between e.g. an unrefined facet and a non-existent facet.

- `SearchResults.getFacetValues` #720
- `SearchResults.getFacetStats` #721
- `SearchParameters.getConjunctiveRefinements` #724
- `SearchParameters.getDisjunctiveRefinements` #725
- `SearchParameters.getExcludeRefinements` #726
- `SearchParameters.isFacetRefined` #727
- `SearchParameters.isExcludeRefined` #728
- `SearchParameters.isDisjunctiveFacetRefined` #729
- `SearchParameters.isHierarchicalFacetRefined` #730
- `SearchParameters.getHierarchicalFacetBreadcrumb` #723

We are keeping:

- `SearchParameters.addXXX`
- `SearchParameters.removeXXX`
- `SearchParameters.toggleXXX`

Other:

- `helper.isRefined` will simply be removed because it's deprecated #731

Changelog:

#732

PR #723, opened 2019-07-10
fix(getHierarchicalFacetBreadcrumb): don't throw an error

Haroenv:
Funnily enough, this test didn't fail becuse the bind always threw

We return an empty array if the facet wasn't declared

see #722

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -1443,6 +1443,10 @@ SearchParameters.prototype = {
    * @return {array.<string>} the path as an array of string
    */
   getHierarchicalFacetBreadcrumb: function(facetName) {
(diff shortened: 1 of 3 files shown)


src/SearchParameters/index.js
```diff
@@ -1444,8 +1444,7 @@ SearchParameters.prototype = {
    */
   getHierarchicalFacetBreadcrumb: function(facetName) {
     if (!this.isHierarchicalFacet(facetName)) {
-      throw new Error(
-        'Cannot get the breadcrumb of an unknown hierarchical facet: `' + facetName + '`');
+      return [];
```
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/726#discussion_r302110392

PR #724, opened 2019-07-10
fix(getConjunctiveRefinements): no error when requested facet is not conjunctive

Haroenv:
Instead we will return [] there.

see #722


src/SearchParameters/index.js
```diff
@@ -536,7 +536,7 @@ SearchParameters.prototype = {
    */
   getConjunctiveRefinements: function(facetName) {
     if (!this.isConjunctiveFacet(facetName)) {
-      throw new Error(facetName + ' is not defined in the facets attribute of the helper configuration');
+      return [];
```
samouss: https://github.com/algolia/algoliasearch-helper-js/pull/726#discussion_r302110392

PR #726, opened 2019-07-10
fix(getExcludeRefinements): replace error by default value

Haroenv:
There was no test for this yet

see #722

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -570,7 +570,7 @@ SearchParameters.prototype = {
    */
   getExcludeRefinements: function(facetName) {
     if (!this.isConjunctiveFacet(facetName)) {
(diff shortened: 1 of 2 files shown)


src/SearchParameters/index.js
```diff
@@ -570,7 +570,7 @@ SearchParameters.prototype = {
    */
   getExcludeRefinements: function(facetName) {
     if (!this.isConjunctiveFacet(facetName)) {
-      throw new Error(facetName + ' is not defined in the facets attribute of the helper configuration');
+      return [];
```
samouss: Not sure the branch is still useful. The return value already fallback on `[]`.

Haroenv: not sure, but this is the most consistent way, all getXXXrefinemnent will use the fallback if it's not declared

Haroenv:
the typings indeed are already correct here

PR #728, opened 2019-07-10
fix(isExcludeRefined): remove error in favor of false

Haroenv:
same as #727, but changed the names

see #722


Haroenv:
looks like I need to fix a different test first

PR #729, opened 2019-07-10
fix(isDisjunctiveFacetRefined): return false if not in disjunctiveFacets

Haroenv:
same as #727 again

see #722


test/spec/SearchParameters/isDisjunctiveFacetRefined.js
```diff
@@ -0,0 +1,46 @@
...
+    disjunctiveFacets: ['test'],
+    disjunctiveFacetsRefinements: {
+      test: ['zongo']
+    }
+  });
+
+  expect(state.isDisjunctiveFacetRefined('test', 'not zongo')).toBe(false);
+});
+
+test('isDisjunctiveFacetRefined returns false if facet is not conjunctive', function() {
```
Haroenv:
```suggestion
test('isDisjunctiveFacetRefined returns false if facet is not disjunctive', function() {
```

PR #730, opened 2019-07-10
fix(isHierarchicalFacetRefined): return false if refinement isn't a facet

Haroenv:
Similar to #727, except here we _do_ need to keep throwing the error on add

see #722

### DIFF
--- a/src/SearchParameters/index.js
+++ b/src/SearchParameters/index.js
@@ -1066,6 +1066,9 @@ SearchParameters.prototype = {
     if (this.isHierarchicalFacetRefined(facet)) {
       throw new Error(facet + ' is already refined.');
     }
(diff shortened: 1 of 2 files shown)


src/SearchParameters/index.js
```diff
@@ -1066,6 +1066,9 @@ SearchParameters.prototype = {
     if (this.isHierarchicalFacetRefined(facet)) {
       throw new Error(facet + ' is already refined.');
     }
+    if (!this.isHierarchicalFacet(facet)) {
+      throw new Error(facet + ' is not defined in the hierarchicalFacets attribute of the helper configuration.');
```
samouss: We could add a test to cover this case.

Haroenv: there's already a test to cover this (the one which failed in the first commit)

PR #737, opened 2019-07-18
feat(requestBuilder): prevent needless extra requests for empty refinements

Haroenv:
We are now moving towards passing empty arrays to refinements; so we need to make sure no extra requests are done.

### DIFF
--- a/test/spec/requestBuilder.js
+++ b/test/spec/requestBuilder.js
@@ -64,5 +64,5 @@ test('does only a single query if refinements are empty', function() {
   });
 
   var queries = getQueries(searchParams.index, searchParams);
-  expect(queries.length).toBe(1);
+  expect(queries).toHaveLength(1);
 });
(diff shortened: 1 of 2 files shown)


test/spec/requestBuilder.js
```diff
@@ -46,3 +47,22 @@ test('The request builder should should force clickAnalytics to false on subsequ
...
+    numericRefinements: {
+      test_numeric: {}
+    },
+    hierarchicalFacetsRefinements: {
+      test_hierarchical: []
+    }
+  });
+
+  var queries = getQueries(searchParams.index, searchParams);
+  expect(queries.length).toBe(1);
```
Haroenv: was 4 on `next`

samouss:
```suggestion
  expect(queries).toHaveLength(1);
```

PR #738, opened 2019-07-22
fix(toggleRefinement): keep an empty array when clearing

Haroenv:
When clearing a refinement using `toggleRefinement` we want to keep the empty refinement object for that attribute, because we use it to override a deeper refinement of that attribute in IS.js with federated search

### DIFF
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -52,7 +52,9 @@ var lib = {
    */
   removeRefinement: function removeRefinement(refinementList, attribute, value) {
     if (value === undefined) {
-      return lib.clearRefinement(refinementList, attribute);
+      return lib.clearRefinement(refinementList, function(v, f) {
+        return attribute === f;
+      });
     }
 
     var valueAsString = '' + value;
--- a/src/SearchParameters/RefinementList.js
+++ b/src/SearchParameters/RefinementList.js
@@ -107,14 +107,11 @@ var lib = {
         var facetList = values.filter(function(value) {
           return !attribute(value, key, refinementType);
         });
-        if (facetList.length > 0) {
-          if (facetList.length !== values.length) {
-            hasChanged = true;
-          }
-          memo[key] = facetList;
-        } else {
+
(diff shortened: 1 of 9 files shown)


test/spec/SearchParameters/removeXFacetRefinement.js
```diff
@@ -0,0 +1,304 @@
...
+        }
+      }
+    });
+
+    expect(state.removeNumericRefinement('attribute')).toEqual(
+      new SearchParameters({
+        numericRefinements: {
+          attribute: {
+            '>=': []
+          }
```
Haroenv: This might be unexpected since there’s no “removeNumericFacet” to remove it fully

Haroenv: This works as expected, since we exclude empty operators in InstantSearch in the getRefinements reimplementation there

src/SearchParameters/RefinementList.js
```diff
@@ -52,7 +52,9 @@ var lib = {
    */
   removeRefinement: function removeRefinement(refinementList, attribute, value) {
     if (value === undefined) {
-      return lib.clearRefinement(refinementList, attribute);
+      return lib.clearRefinement(refinementList, function(v, f) {
```
samouss: We can add a comment that explain why we force the usage of the function.

test/spec/SearchParameters/removeXFacetRefinement.js
samouss: We don't have a function `removeNumericFacet` but it's our responsibility to remove omit it inside `dispose`.

Haroenv:
tried it in #3976, and it did cause some expected test failures in non-migrated widgets, but fixed the issue for refinementlist

src/SearchParameters/RefinementList.js
Haroenv:
```suggestion
      // we use the "filter" form of clearRefinement, since it leaves empty values as-is
      // the form with a string will remove the attribute completely
      return lib.clearRefinement(refinementList, function(v, f) {
```

PR #348, opened 2016-08-17
feat(SearchResults): backport instantsearch.js getRefinement method

bobylito:
Fixes #195

PR #741, opened 2019-08-01
fix(searchOnce): only respond with SearchResults

Haroenv:
State and raw results are already available from a SearchResults object.

the _originalResponse is really odd, it was added without comment in https://github.com/algolia/algoliasearch-helper-js/pull/348/commits/e3a323840484d6a594edb7bc8eb8c01003d38b11#diff-cbf795d639d5f4b610b6d0f6ee395b2aR163 which seems unrelated to me.

TODO:
- [ ] mention in a changelog
- [ ] update integration test

This needs to be merged in a major version, so is blocked for now


Haroenv:
we don't require searchOnce anymore, maybe in a next major version it should even be removed?

Haroenv:
actually it makes sense to do this, if we decide to not remove it in a next major?

PR #743, opened 2019-08-07
fix(removeXFacet): make sure this fully removes empty arrays

Haroenv:
This does not have an averse on the toggle methods, which should keep empty arrays.

required for https://github.com/algolia/instantsearch.js/pull/4024

### DIFF
--- a/test/spec/SearchParameters/removeXFacet.js
+++ b/test/spec/SearchParameters/removeXFacet.js
@@ -1,0 +1,138 @@
+'use strict';
+
+var SearchParameters = require('../../../src/SearchParameters');
+
+describe('removeDisjunctiveFacet', function() {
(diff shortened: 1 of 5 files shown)


test/spec/SearchParameters/removeXFacet.js
```diff
@@ -0,0 +1,138 @@
+'use strict';
+
+var SearchParameters = require('../../../src/SearchParameters');
+
+describe('removeDisjunctiveFacet', function() {
```
francoischalifour: We can perhaps add a test with the default `state = new SearchParameters()`.

Haroenv: added empty parameters tests

PR #754, opened 2019-12-02
fix(types): add `optionalFilters` search parameter

francoischalifour:
This adds the missing [`optionalFilters`](https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/) search parameter type.


francoischalifour:
@Haroenv No, it's not either.

PR #755, opened 2020-01-07
[WIP] chore: migrates to js v4

nunomaduro:


package.json
```diff
@@ -95,6 +95,6 @@
     "qs": "^6.5.1"
   },
   "peerDependencies": {
-    "algoliasearch": ">= 3.1 < 4"
+    "algoliasearch": ">= 4.0"
```
Haroenv: changing this would be a major version, doesn't the API still accept both v3 & v4 in compat mode??

nunomaduro: Let's discuss this tomorrow in our meeting.

PR #756, opened 2020-01-13
feat(algoliasearch): add support for algoliasearch v4 the helper v3

yannickcr:
This change add support for algoliasearch v4 to algoliasearch helper.

No changes were necessary in the code. The test suite was updated to support both v3 and v4 versions of algoliasearch client. We also added a new job to CircleCI to test the version 4 of the client.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,8 @@
+// Note: Bellow, we will be importing both algoliasearch
+// `v3` and algoliasearch `v4` types. The goal is being
+// able to export the algoliasearch-helper types using
+// the developer installed version of the client.
+
 import algoliasearch, {
   // @ts-ignore
   SearchClient as SearchClientV4,
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,10 +8,10 @@ import algoliasearch, {
   // @ts-ignore
   Response as SearchResponseV3,
 } from 'algoliasearch';
-// @ts-ignore
 import {
   SearchOptions as SearchOptionsV4,
-  SearchResponse as SearchResponseV4,
+  SearchResponse as SearchResponseV4
+// @ts-ignore
 } from '@algolia/client-search';
 import { EventEmitter } from 'events';
 
@@ -1048,8 +1048,65 @@ declare namespace algoliasearchHelper {
     type Operator = '=' | '>' | '>=' | '<' | '<=' | '!=';
   }
 
-  export interface SearchResults<T = any>
-    extends Omit<SearchResponse<T>, 'facets' | 'params'> {
+  export class SearchResults<T = any>
+    implements Omit<SearchResponse<T>, 'facets' | 'params'> {
+    /**
+     * query used to generate the results
+     */
+    query: string;
+    /**
+     * The query as parsed by the engine given all the rules.
+     */
+    parsedQuery: string;
+    /**
+     * all the records that match the search parameters. Each record is
+     * augmented with a new attribute `_highlightResult`
+     * which is an object keyed by attribute and with the following properties:
+     *  - `value` : the value of the facet highlighted (html)
+     *  - `matchLevel`: full, partial or none depending on how the query terms match
+     */
+    hits: (T & {
+      readonly objectID: string;
+    })[];
+    /**
+     * index where the results come from
+     */
+    index: string;
+    /**
+     * number of hits per page requested
+     */
+    hitsPerPage: number;
+    /**
+     * total number of hits of this query on the index
+     */
+    nbHits: number;
+    /**
+     * total number of pages with respect to the number of hits per page and the total number of hits
+     */
+    nbPages: number;
+    /**
+     * current page
+     */
+    page: number;
+    /**
+     * sum of the processing time of all the queries
+     */
+    processingTimeMS: number;
+    /**
+     * The position if the position was guessed by IP.
+     * @example "48.8637,2.3615",
+     */
+    aroundLatLng: string;
+    /**
+     * The radius computed by Algolia.
+     * @example "126792922",
+     */
+    automaticRadius: string;
+    /**
+     * String identifying the server used to serve this request.
+     * @example "c7-use-2.algolia.net",
+     */
+    serverUsed: string;
     /**
      * Boolean that indicates if the computation of the counts did time out.
      * @deprecated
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@types/algoliasearch": "^3.30.12",
     "algolia-frontend-components": "0.0.35",
-    "algoliasearch": "3.32.0",
+    "algoliasearch": ">= 3.32.0 < 5",
```
francoischalifour: The dev dependency should be pinned to `4.0.0-beta.14`.

package.json
```diff
@@ -98,6 +98,6 @@
     "events": "^1.1.1"
   },
   "peerDependencies": {
-    "algoliasearch": ">= 3.1 < 4"
+    "algoliasearch": ">= 3.32.0 < 5"
```
francoischalifour: Did you notice incompatibilities with `algoliasearch` < 3.32.0?

nunomaduro: The `algoliasearch-helper-js` is not compatible with versions prior to `3.32.0`.

package.json
nunomaduro: Why buddy?

francoischalifour: For consistency. We always want a reproducible development environment, which is why we pin all dev dependencies.

You can read more on the [Renovate website](https://docs.renovatebot.com/dependency-pinning/).

package.json
Haroenv: What's broken before 3.32? Narrowing the peer dependency is a major version IIRC

nunomaduro:
@Haroenv @francoischalifour Just updated the pull request addressing your comments. Changes: https://github.com/algolia/algoliasearch-helper-js/pull/756/commits/fe3e3944a3e42488183ab5f69f5aaa0863a0217f.

@Haroenv For info, the tests won't pass with the version `3.31` of algoliasearch because the userAgent wasn't computed in the same way.

test/spec/algoliasearch.helper/client.js
```diff
@@ -58,11 +64,17 @@ test('clearCache gets called if exists', function() {
...
+      return client.transporter.userAgent.value;
+    }
+  });
+
   var originalUA = client._ua;
   algoliaSearchHelper(client, 'IndexName', {});
   algoliaSearchHelper(client, 'IndexName2', {});
 
-  expect(client._ua).toBe(originalUA + ';JS Helper (' + version + ')');
+  expect(client._ua).toBe(originalUA + '; JS Helper (' + version + ')');
```
francoischalifour: Did you check with the Data team that these extra spaces don't cause trouble with their aggregation?

nunomaduro: No problem on that. It's already like this since version `3.33`. We are updating now, because previously we were testing against `3.32`: https://github.com/algolia/algoliasearch-helper-js/pull/756/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L43.

Now we are testing against `3.35.1`: https://github.com/algolia/algoliasearch-helper-js/pull/756/files#diff-1d37e48f9ceff6d8030570cd36286a61R67

Haroenv:
I understand what you mean, but that means that only the tests don't pass, not that you can't use it with an older client. Really the problem is that algoliasearch isn't a peer dependency in the strict sense of the word, since we don't use it directly :(

francoischalifour:
@Haroenv @nunomaduro removed the range change 👍

nunomaduro:
@Haroenv Got it. Here is the update: https://github.com/algolia/algoliasearch-helper-js/pull/756/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R101. Now is good.

index.d.ts
```diff
@@ -1,8 +1,28 @@
-import algoliasearch from 'algoliasearch';
+import algoliasearch, {
+  // @ts-ignore
```
Haroenv: can we put a comment before these imports, like:

```
// we are importing both v3 and v4 types
// they  will  be `any` respectively, which will allow us  to migrate
// cleanly
```

index.d.ts
```diff
@@ -1,8 +1,28 @@
...
+  SearchOptions as SearchOptionsV4,
+  SearchResponse as SearchResponseV4
+} from '@algolia/client-search';
 import { EventEmitter } from 'events';
 
+type DummySearchClientV4 = {readonly addAlgoliaAgent: (segment: string, version?: string) => void};
+
+type Client = ReturnType<typeof algoliasearch> extends DummySearchClientV4 ? SearchClientV4 : SearchClientV3
+type SearchOptions = ReturnType<typeof algoliasearch> extends DummySearchClientV4 ? SearchOptionsV4 : SearchOptionsV3
+type SearchResponse<T> = ReturnType<typeof algoliasearch> extends DummySearchClientV4 ? SearchResponseV4<T> : SearchResponseV3<T>
```
Haroenv: let's run prettier here

index.d.ts
```diff
@@ -1019,63 +1039,7 @@ declare namespace algoliasearchHelper {
...
-     * The radius computed by Algolia.
-     * @example "126792922",
-     */
-    automaticRadius: string;
-    /**
-     * String identifying the server used to serve this request.
-     * @example "c7-use-2.algolia.net",
-     */
-    serverUsed: string;
+  export interface SearchResults<T = any> extends Omit<SearchResponse<T>, 'facets' | 'params'> {
```
Haroenv: IIRC it needs to be a class, since we are instantiating it

index.d.ts
```diff
@@ -1019,63 +1048,8 @@ declare namespace algoliasearchHelper {
...
-     * The radius computed by Algolia.
-     * @example "126792922",
-     */
-    automaticRadius: string;
-    /**
-     * String identifying the server used to serve this request.
-     * @example "c7-use-2.algolia.net",
-     */
-    serverUsed: string;
+  export interface SearchResults<T = any>
```
Haroenv: this still needs to be a class, so we might need to have this duplication

index.d.ts
```diff
@@ -1,10 +1,44 @@
...
-  'search' | 'searchForFacetValues'
->;
+type DummySearchClientV4 = {
+  readonly addAlgoliaAgent: (segment: string, version?: string) => void;
+};
+
+type Client = ReturnType<typeof algoliasearch> extends DummySearchClientV4
+  ? SearchClientV4
+  : SearchClientV3;
+type SearchOptions = ReturnType<
```
francoischalifour: There's no way to reuse the `Client` type?

Something like:

```ts
type SearchOptions = Client extends SearchClientV4
  ? SearchOptionsV4
  : SearchOptionsV3;
```

nunomaduro: `SearchClientV4` may be any when is not installed. Everything extends any. So in your example `SearchOptions` will be any too.

Haroenv: I think @francoischalifour meant: `type SearchOptions = Client extends DummySearchClientV4`

PR #762, opened 2020-04-02
fix(defaultsPure): don't change keys order, fix #761

### LINKED ISSUE #761: Unable to disable facet's values filtering since v3
Hi :wave: 

I think I found a bug with `defaultsPure()` when using `.getFacetValues()` when sorting is disabled:
```js
event.results.getFacetValues('offer.name', {
  sortBy: function(a, b) {
    return 0; // disable sorting
  }
});
```

Imagine five values for our filter `offer.name`. 
Sélection_247

I select `OFFER_TYPE_PRO`, but this is what happens with algoliasearch-helper-js v3:
Sélection_248

It takes the 1st position where it shouldn't, it should stay in the 3rd position.

This is what happens using algoliasearch-helper-js v2, the expected result:
Sélection_249

The issue is not directly comming from `getFacetValues()` and the way it sort values, but from its source of truth: `SearchResults#disjunctiveValues`.

Sélection_246

For info, this is the Algolia response, you can see `OFFER_TYPE_PRO` is still in 3rd position for facet `offer.name`:
Sélection_250

This is a part of `SearchResults` constructor:
- `facetResults` is the same than Algolia response, everything is okay
- `dataFromMainRequest` is the value we selected just before, perfect
- but when it comes to `defaultsPure()` our selected value is put back at top of values, this is not wanted
Sélection_244

I think this issue can only happens when we disable facet values sorting.
By default it's sorted by alpha/count and everything is okay, but it's not when we don't want to sort values.

I will try to work on a PR.

Kocal:
See #761 for issue, thanks!

I'm computing an array of keys in the order they come from sources, then using it with `mergedSources` to return a new _keys-order-preserved_ object.

### DIFF
--- a/src/functions/defaultsPure.js
+++ b/src/functions/defaultsPure.js
@@ -14,12 +14,7 @@ module.exports = function defaultsPure() {
   }, {});
 
   var orderedKeys = sources.reduce(function(acc, source) {
-    Object.keys(Object(source)).forEach(function(key) {
-      if (!acc.includes(key)) {
-        acc.push(key);
-      }
-    });
-    return acc;
+    return acc.slice().concat(Object.keys(Object(source)));
   }, []);
 
   return orderedKeys.reduce(function(acc, key) {
--- a/src/functions/defaultsPure.js
+++ b/src/functions/defaultsPure.js
@@ -1,14 +1,30 @@
 'use strict';
 
 // NOTE: this behaves like lodash/defaults, but doesn't mutate the target
+// it also preserve keys order
 module.exports = function defaultsPure() {
   var sources = Array.prototype.slice.call(arguments);
-  return sources.reduceRight(function(acc, source) {
+  var mergedSources = sources.reduceRight(function(acc, source) {
     Object.keys(Object(source)).forEach(function(key) {
       if (source[key] !== undefined) {
         acc[key] = source[key];
       }
     });
+    return acc;
+  }, {});
+
+  var orderedKeys = sources.reduce(function(acc, source) {
+    Object.keys(Object(source)).forEach(function(key) {
+      if (!acc.includes(key)) {
+        acc.push(key);
+      }
+    });
+    return acc;
+  }, []);
+
+  return orderedKeys.reduce(function(acc, key) {
+    acc[key] = mergedSources[key];
+
     return acc;
   }, {});
 };
--- a/src/functions/defaultsPure.js
+++ b/src/functions/defaultsPure.js
@@ -4,22 +4,12 @@
 // it also preserve keys order
 module.exports = function defaultsPure() {
   var sources = Array.prototype.slice.call(arguments);
-  var mergedSources = sources.reduceRight(function(acc, source) {
+  return sources.reduce(function(acc, source) {
     Object.keys(Object(source)).forEach(function(key) {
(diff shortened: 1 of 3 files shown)


src/functions/defaultsPure.js
```diff
@@ -1,14 +1,30 @@
...
       if (source[key] !== undefined) {
         acc[key] = source[key];
       }
     });
     return acc;
   }, {});
+
+  var orderedKeys = sources.reduce(function(acc, source) {
+    Object.keys(Object(source)).forEach(function(key) {
+      if (!acc.includes(key)) {
```
Haroenv, line 18:
```suggestion
      if (acc.indexOf(key) === -1) {
```

Haroenv:
This seems interesting, but I wonder if there's not a way where there's less iterations, this is a quite hot function

Kocal:
I think we can, I've replaced the `var orderedKeys = ...` by:
```js
  var orderedKeys = sources.reduce(function(acc, source) {
    return [...acc, ...Object.keys(source)];
  }, []);
```

eunjae-lee:
I've got a question. 

```
defaults({'a': 1, 'b': 2}, {'b': 3, 'c': 3}, {'d': 4});
```

What is the correct order of the keys in the example above?

Do we allow such case? or do we always assume the first source has all the keys?

Kocal:
For me, the keys order should be `a, b, c, d`.

> Do we allow such case? or do we always assume the first source has all the keys?

Keys are taken from all sources, from left to right, keys are added only if they are not already present.

But since https://github.com/algolia/algoliasearch-helper-js/pull/762#issuecomment-607791799, I looks like to me it won't work as expected. I prefer this solution https://github.com/algolia/algoliasearch-helper-js/pull/762/commits/2341a8192ca9c80e637bda2326125ea883a504cc#diff-71f496a2aca33d97c37bacd684267742R16-R23

Kocal:
Hi, is there any updates on this? 

Thanks

Haroenv:
Sorry, this function is used in many places, and these three more iterations will have a negative impact, which will also be the case using concat still.

For me it's a little unclear how these changes fix the issue you're describing. Do you think you could add a test case which is more like the exact case you want fixed? Thanks!

yannickcr:
Proposing this implementation:

```js
module.exports = function defaultsPure() {
  var sources = Array.prototype.slice.call(arguments);
  return sources.reduce(function(acc, source) {
    Object.keys(Object(source)).forEach(function(key) {
      if (source[key] !== undefined && !Object.hasOwnProperty.call(acc, key)) {
        acc[key] = source[key];
      }
    });
    return acc;
  }, {});
};
```

Its performance should be very close to the current implementation (the only changes being the use of `reduce` instead of `reduceRight` and the additional check with `Object.hasOwnProperty`) while keeping the key order on the final object.

WDYT @Kocal @Haroenv ?

Kocal:
Hi @yannickcr, thanks for your proposal!

I'm a big fan of your implementation, tests still pass and as you say performance should be very close to the current implementation.

Kocal:
I've added a failing test (935969ca0edcc09d5fb5f9ac9fd5b33f3e00cb58), it will fails if you revert modification on `defaultsPure`.

PR #775, opened 2020-06-01
fix(typescript): fix TypeScript 3.9 compatibility (fixes #774)

### LINKED ISSUE #774: Types Error:  An interface can only extend an object type or intersection of object types with statically known members
We see the following type error with TS 3.9 and 4

```
node_modules/algoliasearch-helper/index.d.ts(348,50):
TS2312: An interface can only extend an object type or
 intersection of object types with statically known members.
```
---

https://github.com/algolia/algoliasearch-helper-js/blob/develop/index.d.ts#L348

```
type SearchOptions = Client extends DummySearchClientV4
  ? SearchOptionsV4
  : SearchOptionsV3;

export interface PlainSearchParameters extends SearchOptions {}
```

> Interfaces can't extend conditional types because we can't know that all possible endpoints of a conditional type result in something that's legal to extend an interface from. For example, you can write interface I extends number { but you might have a conditional type that for some type resolves to number.
 [Source](https://github.com/microsoft/TypeScript/issues/23850#issuecomment-386194033)

yannickcr:
We are doing some tricky things here by using [Conditional Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types) to keep this library compatible with [`algoliasearch`](https://github.com/algolia/algoliasearch-client-javascript) v3 and v4.

But since 3.9 TypeScript is not able to statically analyze our code anymore to determine the `SearchOptions` and `SearchResponse` types (https://github.com/algolia/algoliasearch-helper-js/issues/774) :sob: 

This PR fixes this by using directly `ReturnType<typeof algoliasearch>` instead of the `Client` type. With this notation the static analyze seems to work with no issues again.

fixes #774

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -27,10 +27,10 @@ type DummySearchClientV4 = {
 type Client = ReturnType<typeof algoliasearch> extends DummySearchClientV4
   ? SearchClientV4
   : SearchClientV3;
-type SearchOptions = Client extends DummySearchClientV4
+type SearchOptions = ReturnType<typeof algoliasearch> extends DummySearchClientV4
   ? SearchOptionsV4
   : SearchOptionsV3;
-type SearchResponse<T> = Client extends DummySearchClientV4
+type SearchResponse<T> = ReturnType<typeof algoliasearch> extends DummySearchClientV4
   ? SearchResponseV4<T>
   : SearchResponseV3<T>;
 


yannickcr:
> this is pretty weird

Yeah, I think we are playing very close from the static analyses limitations of TS here :neutral_face:

Haroenv:
This trick doesn't seem to work with TS 3.7 🤔

PR #779, opened 2020-06-23
fix: accept all fields implicitly instead of the allow list

eunjae-lee:
This PR updates SearchResults to expose all the fields from the search API response implicitly.
It used to manage an allow-list. I've removed it and only computed properties are added on top of that. It means if search API adds a new field, it might be overridden by the computed property if their names are the same.

- [x] I've removed the allow-list with the jsdoc and the doc building is failing. Let me see what to do.

Do you see any improvements that can be made? cc @Haroenv

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -240,11 +240,15 @@ function SearchResults(state, results) {
    * query used to generate the results
    * @name query
    * @member {string}
+   * @memberof SearchResults
+   * @instance
    */
   /**
    * The query as parsed by the engine given all the rules.
    * @name parsedQuery
    * @member {string}
+   * @memberof SearchResults
+   * @instance
    */
   /**
    * all the records that match the search parameters. Each record is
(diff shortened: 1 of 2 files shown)


src/SearchResults/index.js
```diff
@@ -229,50 +229,11 @@ function SearchResults(state, results) {
...
-   * @member {number}
-   */
-  this.nbPages = mainSubResponse.nbPages;
-  /**
-   * current page
-   * @member {number}
-   */
-  this.page = mainSubResponse.page;
+  // https://www.algolia.com/doc/api-reference/api-methods/search/#response
+  Object.keys(mainSubResponse).forEach((key) => {
```
Haroenv, line 233:
```suggestion
  Object.keys(mainSubResponse).forEach(function (key) {
```

eunjae-lee: yeah I just found it out 😅 Jest didn't throw on my computer, but did in CircleCI.

Haroenv:
can you locally diff the doc build on this branch vs. develop, is it different?

test/datasets/SearchParameters/search.dataset.js
```diff
@@ -355,6 +355,7 @@ function getData() {
     'nbHits': 4,
     'nbPages': 1,
     'page': 0,
+    'params': 'query=&hitsPerPage=20&page=0&facets=%5B%5D&facetFilters=%5B%5B%22city%3AParis%22%2C%22city%3ANew%20York%22%5D%5D',
```
eunjae-lee, line 358: Since this PR, `params` is added to SearchResults. Any reason to exclude this? Or let it happen?

Haroenv: I don't think it should be removed, there's not really a reason to read it since there's already the SearchParameters class, so we don't need to worry about people abusing this

Haroenv:
has a .2kB impact 😃 --> 11 -> 10.8

eunjae-lee:
Yeah I just build the doc from `develop` branch and the output is the same.
I had to put `@memberof` and `@instance` for the tweak.

Haroenv:
While I can't see a negative impact, can @yannickcr & @francoischalifour review this too?

PR #786, opened 2020-07-21
fix(defaultsPure): fix the regression where the order was wrong with addFacetRefinement

eunjae-lee:
This PR fixes the regression of #762 where the order became wrong when running `addFacetRefinement` multiple times.

I don't know how to make the implementation of the `defaultsPure` shorter/cleaner/easier, but the bottom line is we need to pass the test cases I've added in this PR.

I've removed a wrong test case added in #762 which is

```js
it('should not touch keys order', function() {
  var expected = {'a': 1, 'b': 3, 'c': 3, 'd': 4};
  var actual = defaults({'a': 1, 'b': 2}, {'b': 3, 'c': 3}, {'d': 4});
```

It's a conflict and it doesn't make sense with our current use cases.

### DIFF
--- a/src/functions/defaultsPure.js
+++ b/src/functions/defaultsPure.js
@@ -2,36 +2,32 @@
 
 // NOTE: this behaves like lodash/defaults, but doesn't mutate the target
 // it also preserve keys order
+var PLACEHOLDER = '__PLACEHOLDER__';
+
 module.exports = function defaultsPure() {
   var sources = Array.prototype.slice.call(arguments);
 
-  var keysInOrder = Object.keys(sources.reduceRight(function(acc, source) {
+  var emptyObjectWithKeysInOrder = sources.reduceRight(function(acc, source) {
     Object.keys(Object(source)).forEach(function(key) {
       if (source[key] === undefined) {
         return;
       }
-      if (acc[key] === null) {
+      if (acc[key] === PLACEHOLDER) {
         // remove if already added, so that we can add it in correct order
         delete acc[key];
       }
-      acc[key] = null;  // placeholder
+      acc[key] = PLACEHOLDER;
     });
     return acc;
-  }, {}));
+  }, {});
 
-  var merged = sources.reduce(function(acc, source) {
+  return sources.reduce(function(acc, source) {
     Object.keys(Object(source)).forEach(function(key) {
-      if (source[key] !== undefined && !Object.hasOwnProperty.call(acc, key)) {
+      if (source[key] !== undefined && acc[key] === PLACEHOLDER) {
         acc[key] = source[key];
       }
     });
     return acc;
-  }, {});
-
-  var final = keysInOrder.reduce(function(acc, key) {
-    acc[key] = merged[key];
-    return acc;
-  }, {});
-
-  return final;
+  }, emptyObjectWithKeysInOrder);
 };
+
--- a/src/functions/defaultsPure.js
+++ b/src/functions/defaultsPure.js
@@ -2,32 +2,20 @@
 
 // NOTE: this behaves like lodash/defaults, but doesn't mutate the target
 // it also preserve keys order
-var PLACEHOLDER = '__PLACEHOLDER__';
-
 module.exports = function defaultsPure() {
   var sources = Array.prototype.slice.call(arguments);
 
-  var emptyObjectWithKeysInOrder = sources.reduceRight(function(acc, source) {
+  return sources.reduceRight(function(acc, source) {
     Object.keys(Object(source)).forEach(function(key) {
       if (source[key] === undefined) {
         return;
       }
-      if (acc[key] === PLACEHOLDER) {
+      if (acc[key] !== undefined) {
         // remove if already added, so that we can add it in correct order
         delete acc[key];
       }
-      acc[key] = PLACEHOLDER;
+      acc[key] = source[key];
     });
     return acc;
   }, {});
-
-  return sources.reduce(function(acc, source) {
-    Object.keys(Object(source)).forEach(function(key) {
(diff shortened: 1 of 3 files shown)


Haroenv:
could you test the performance impact and compare that to both V3.1.1 & 3.1.2? I'm interested in seeing what impact this has (since it's a hot function)

eunjae-lee:
> could you test the performance impact and compare that to both V3.1.1 & 3.1.2? I'm interested in seeing what impact this has (since it's a hot function)

The time complexity is the same O(N), but anyway the implementation is this PR runs the loop three times, compared to one time in previous versions. I will try to find a way to test the performance tomorrow.

Haroenv:
I just have an overall question, what does the order of facets matter for?

eunjae-lee:
> I just have an overall question, what does the order of facets matter for?

In our connectors, for example, in [connectCurrentRefinements](https://github.com/algolia/instantsearch.js/blob/master/src/connectors/current-refinements/connectCurrentRefinements.ts#L258-L260) and [connectRefinementList](https://github.com/algolia/instantsearch.js/blob/master/src/connectors/refinement-list/connectRefinementList.js#L336:L336), we sort the facets by ourselves. So it's likely to be okay. I haven't checked everything, but even if everything is okay, someone who's using helper directly or disabled sortBy option in the connector, then they'll go through breaking change.

Without this fix, the behavior of [addRefinement](https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/SearchParameters/RefinementList.js#L27-L43) changes which isn't our intention.

eunjae-lee:
I know it's not a sophisticated performance test, but here's what I've done.

```js
it('test performance', () => {
  [1000, 100000].forEach(function (times, index) {
    console.log('Round ' + (index + 1) + ' (' + times + ' times' + ')');
    var t1 = new Date().getTime();
    for (var i = 0; i < times; i++) {
      defaults({'a': 1, 'b': 2}, {'b': 3}, {'c': 3});
    }
    var t2 = new Date().getTime();
    console.log('Test #1 -> ', t2 - t1);

    t1 = new Date().getTime();
    for (i = 0; i < times; i++) {
      defaults({},
        {
          'Insignia™': 551,
          'Samsung': 511,
          'Apple': 386
        },
        {
          'Apple': 386
        });
    }
    t2 = new Date().getTime();
    console.log('Test #2 -> ', t2 - t1);
  })
});
```

and I ran `yarn test:unit -t "test performance"`.


|                      | pr #786 | v3.2.0 | v3.1.2 | v3.1.1 |
|----------------------|---------|--------|--------|--------|
| Round 1 (1k times)   |         |        |        |        |
| Test 1              | 16      | 9      | 9      | 7      |
| Test 2              | 15      | 18     | 9      | 7      |
| Round 2 (100k times) |         |        |        |        |
| Test 1              | 1160    | 661    | 791    | 563    |
| Test 2              | 545     | 425    | 702    | 402    |

As I said yesterday, the more number of items given to `defaultsPure`, the longer it will take because it's O(N), and this PR has more impacts on it.

In Test 2, the first argument is an empty object, which means it's basically performing with two objects in the test. And in the helper, `defaultsPure()` is called in 7 places, and 6 out of 7 is that case (an empty object and two other objects). The another one case is where it is called with three objects.

So IMO this PR technically slows down the performance, but practically it will be similar. What do you think?

Haroenv:
The maximum value the objects can be is 100 keys each, could you test that case?

eunjae-lee:
> The maximum value the objects can be is 100 keys each, could you test that case?

I've tested that case, and also improve the performance since the last implementation.

Haroenv (approved):

src/functions/defaultsPure.js
```diff
@@ -2,14 +2,32 @@
...
   return sources.reduce(function(acc, source) {
     Object.keys(Object(source)).forEach(function(key) {
-      if (source[key] !== undefined && !Object.hasOwnProperty.call(acc, key)) {
+      if (source[key] !== undefined && acc[key] === PLACEHOLDER) {
         acc[key] = source[key];
       }
     });
     return acc;
-  }, {});
+  }, emptyObjectWithKeysInOrder);
```
yannickcr, line 31: Maybe you can skip the `emptyObjectWithKeysInOrder` step by just checking if the key already exists:

```js
module.exports = function defaultsPure() {
  var sources = Array.prototype.slice.call(arguments);

  return sources.reduceRight(function(acc, source) {
    Object.keys(Object(source)).forEach(function(key) {
      if (source[key] === undefined) {
        return;
      }
      if (acc[key] !== undefined) {
        // remove if already added, so that we can add it in correct order
        delete acc[key];
      }
      acc[key] = source[key];
    });
    return acc;
  }, {});
};
```

WDYT ?

eunjae-lee: To merge the values, we go from left to right. But to preserve the order of keys, we should go from right to left. That's why I had to have the separate step. And without it, I couldn't think of a way to preserve the order just in one reducer by checking existence of keys. Any idea?

yannickcr: Mmm, the test suite is passing with this code. If this is incorrect maybe we are missing a test case here ?

eunjae-lee: Oh you're right. I misunderstood your code. It works fine. Great job!!! https://github.com/algolia/algoliasearch-helper-js/pull/786/commits/7b88a74908a35c2a99531c3a1b9696c3dce254ee

eunjae-lee:
The performance table has been updated @Haroenv @yannickcr

Haroenv (approved):
still some cases where the perf seems to be lower than all the way before, but some in which it's way better, so good for me!

PR #789, opened 2020-07-30
fix(ts): use a dedicated key to determine client version

Haroenv:
Since the type definition of `@types/algoliasearch` also has `addAlgoliaAgent`, it actually took the v4 imports, which turn out to be "any". Since the change in https://github.com/algolia/algoliasearch-helper-js/pull/775 showed this behaviour more clearly, I found a solution to use a dedicated key which only exists in v4 (transporter) to distinguish between the versions

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -22,6 +22,7 @@ import { EventEmitter } from 'events';
 
 type DummySearchClientV4 = {
   readonly addAlgoliaAgent: (segment: string, version?: string) => void;
+  transporter: any;
 };
 
 type Client = ReturnType<typeof algoliasearch> extends DummySearchClientV4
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -22,6 +22,7 @@ import { EventEmitter } from 'events';
 
 type DummySearchClientV4 = {
   readonly addAlgoliaAgent: (segment: string, version?: string) => void;
```
yannickcr, line 24: I that case do we still need to keep `addAlgoliaAgent` ?

Haroenv: good point, it's indeed not needed anymore
```suggestion
```

PR #791, opened 2020-08-18
fix(ts): compat v3 and v4 without ts-ignore

Haroenv:
For https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45699, it becomes illegal to have a ts-ignore in a dependency, without any way to disable it. I've talked to @orta and finally found this as a solution.

Once this PR has been released, it's possible to update the definitelytyped PR and React InstantSearch will finally work with v4 of algoliasearch without manually removing `@types/algoliasearch` and `@types/algoliasearch-helper`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,21 +2,34 @@
 // `v3` and algoliasearch `v4` types. The goal is being
 // able to export the algoliasearch-helper types using
 // the developer installed version of the client.
+declare module 'algoliasearch' {
+  // fake v3 exports
+  interface Client {}
+  interface QueryParameters {}
+  interface Response<T> {}
+
+  // fake v4 exports as interface
+  interface SearchClientV4 extends SearchClientV4Type {}
+}
+
+declare module '@algolia/client-search' {
+  // fake v4 exports as interface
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -2,43 +2,61 @@
...
+  // fake v4 exports as interface
+  interface SearchOptionsV4 extends SearchOptionsV4Type {}
+  interface SearchResponseV4<TObject = {}> extends SearchResponseV4Type {}
+}
 
 import algoliasearch, {
-  // @ts-ignore
-  SearchClient as SearchClientV4,
-  // @ts-ignore
+  SearchClient as SearchClientV4Type,
```
Haroenv, line 22: this doesn't work in v3 after all

index.d.ts
```diff
@@ -2,43 +2,61 @@
...
+  // fake v3 exports
+  interface Client {}
+  interface QueryParameters {}
+  interface Response<T> {}
+
+  // fake v4 exports as interface
+  interface SearchClientV4 extends SearchClientV4Type {}
+}
+
+declare module '@algolia/client-search' {
```
Haroenv, line 15: declare module for a non-dependency seems to still throw `ts(2307)` too

PR #796, opened 2020-10-30
chore(dev): replace github dependency metalsmith-in-place

Haroenv:
1. replace GitHub dependency with npm dependency
2. use patch-package to make the small GitHub changes

### DIFF
--- a/patches/metalsmith-in-place+1.4.4.patch
+++ b/patches/metalsmith-in-place+1.4.4.patch
@@ -1,0 +1,24 @@
+diff --git a/node_modules/metalsmith-in-place/lib/index.js b/node_modules/metalsmith-in-place/lib/index.js
+index 4c0092a..1340556 100644
+--- a/node_modules/metalsmith-in-place/lib/index.js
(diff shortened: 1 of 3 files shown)


patches/metalsmith-in-place+1.4.4.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/node_modules/metalsmith-in-place/lib/index.js b/node_modules/metalsmith-in-place/lib/index.js
```
Haroenv, line 1: you can test that this diff is correct by checking https://codeload.github.com/superwolff/metalsmith-in-place/tar.gz/7cb06e54142b8843f35178ceb5560946ae356049 which still exists today

I looked into updating to a newer version of metalsmith-in-place, but they added a bunch of dependencies that seemed hard to migrate towards

Haroenv:
try this PR out by running `yarn doc`

yannickcr (approved):
The fix seems correct, but could you add a comment somewhere to explain why this patch is necessary ?

PR #800, opened 2020-11-18
fix(ts): correct type for getNumericRefinements

Haroenv:
this actually returns an object, not an array!


shortcuts (approved):
Good catch!

PR #801, opened 2020-12-02
fix(removeNumericRefinement): clear empty refinements

Haroenv:
Previously this seemingly on purpose checked if this attrubte had refinements, but due to the way it is checking this, it will leave `{[attribute]: {}}` in the numeric refinements, which isn't what we want in e.g. a dispose.

related PR: https://github.com/algolia/instantsearch.js/pull/4587

Since this wasn't covered by any sort of test, I think this behaviour was unintentional

### DIFF
--- a/test/spec/SearchParameters/_clearNumericRefinements.js
+++ b/test/spec/SearchParameters/_clearNumericRefinements.js
@@ -7,6 +7,16 @@ test('When removing all numeric refinements of a state without any', function() 
   expect(state._clearNumericRefinements()).toBe(state.numericRefinements);
 });
 
+test('When removing numericRefinements of a specific attribute, which has no refinements', function() {
+  var state = SearchParameters.make({
+    numericRefinements: {
+      size: {}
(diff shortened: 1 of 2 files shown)


test/spec/SearchParameters/_clearNumericRefinements.js
```diff
@@ -7,14 +7,24 @@ test('When removing all numeric refinements of a state without any', function()
...
+
 test('When removing numericRefinements of a specific attribute, and there are no refinements for this attribute', function() {
   var state = SearchParameters.make({
     numericRefinements: {
       price: {'>': [300]}
     }
   });
 
-  expect(state._clearNumericRefinements('size')).toBe(state.numericRefinements);
+  expect(state._clearNumericRefinements('size')).toEqual(state.numericRefinements);
```
Haroenv, line 27: the values are the same, but omit returns a new object :)

francoischalifour (approved):

PR #802, opened 2020-12-02
fix: ignore invalid userToken

eunjae-lee:
This PR ignores invalid userToken and prints a warning message.


eunjae-lee:
What do you think about this way? I'm looking into why the build is failing, btw.

src/algoliasearch.helper.js
```diff
@@ -886,6 +889,12 @@ AlgoliaSearchHelper.prototype.setIndex = function(name) {
  * helper.setQueryParameter('hitsPerPage', 20).search();
  */
 AlgoliaSearchHelper.prototype.setQueryParameter = function(parameter, value) {
```
Haroenv, line 891: it's not the right place here, people can also call setQueryParameters, but both this and setQueryParameters use setState

src/algoliasearch.helper.js
```diff
@@ -902,6 +911,11 @@ AlgoliaSearchHelper.prototype.setQueryParameter = function(parameter, value) {
  * @chainable
  */
 AlgoliaSearchHelper.prototype.setState = function(newState) {
+  if (newState.userToken !== undefined && !isValidUserToken(newState.userToken)) {
+    delete newState.userToken;
```
Haroenv, line 915: we should leave the invalid token IMO

src/utils/logger.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+var PREFIX = '[algoliasearch-helper]';
+
+function warn(message) {
+  console.warn(PREFIX + ' ' + message);
```
Haroenv, line 6: I don't see an eslint error here, why not using console.warn directly in the setState?

src/SearchParameters/index.js
```diff
@@ -93,6 +95,10 @@ function findArray(array, searchedValue) {
 function SearchParameters(newParameters) {
   var params = newParameters ? SearchParameters._parseNumbers(newParameters) : {};
 
+  if (params.userToken !== undefined && !isValidUserToken(params.userToken)) {
+    delete params.userToken;
+    logger.warn('The `userToken` parameter is invalid. Format: [a-zA-Z0-9_-]{1,64}');
+  }
```
Haroenv, line 101: I think if it's only put in the .make method, both this and .make call it? (or is it the other way round and make calls the constructor?) Either way, I think setState calls make which calls the constructor, so you get three warnings if you do setQueryParameter?

src/utils/logger.js
```diff
@@ -0,0 +1,11 @@
...
+}
+
+module.exports = {
+  warn
```
Haroenv, line 10:
```suggestion
  warn: warn
```

this is causing the build error

eunjae-lee: oh.. I've forgotten about this for a long time.

src/algoliasearch.helper.js
eunjae-lee: oh right, so if I understand correctly, everything leads to here: https://github.com/algolia/algoliasearch-helper-js/blob/fix/invalid-usertoken/src/SearchParameters/index.js#L1369:L1369

src/SearchParameters/index.js
yannickcr: From what I saw, putting the warning it in the constructor should be enough.

We could also add a flag to only display the warning once.

src/SearchParameters/index.js
```diff
@@ -93,6 +95,10 @@ function findArray(array, searchedValue) {
 function SearchParameters(newParameters) {
   var params = newParameters ? SearchParameters._parseNumbers(newParameters) : {};
 
+  if (params.userToken !== undefined && !isValidUserToken(params.userToken)) {
+    delete params.userToken;
+    logger.warn('The `userToken` parameter is invalid. Format: [a-zA-Z0-9_-]{1,64}');
```
yannickcr, line 100: Since it is just a warning and it can have no apparent impact on the app behavior at first, I would add more information to the error message to precise what are the impact of having an invalid `userToken`, with a link to the documentation if possible.

test/spec/algoliasearch.helper/setQueryParameter.js
```diff
@@ -35,3 +35,28 @@ test('setChange should not change the current state: no real modification', func
   expect(changed).toBe(false);
   expect(helper.state).toBe(initialState);
 });
+
+test('setChange should ignore invalid userToken', function() {
```
yannickcr, line 39: We should also test that the warning was displayed if the token is invalid.

test/spec/algoliasearch.helper/setQueryParameter.js
```diff
@@ -35,3 +35,28 @@ test('setChange should not change the current state: no real modification', func
...
+  expect(helper.state.userToken).toBeUndefined();
+
+  helper.setQueryParameter('userToken', '');
+  expect(helper.state.userToken).toBeUndefined();
+
+  helper.setQueryParameter('userToken', 'my invalid token!');
+  expect(helper.state.userToken).toBeUndefined();
+});
+
+test('setChange should preserve valid userToken', function() {
```
Haroenv, line 51: seems like this type exists in the whole file, but, the test should be called `setState`, not `setChange` :)

src/algoliasearch.helper.js
eunjae-lee: @Haroenv 
image
https://excalidraw.com/#json=4883557457592320,c67cypVCpL_-WxoGxVuwKQ

I drew the flow where state can change except for facets (because facet-related methods can never update userToken).

I think everything goes into the constructor of SearchParameters.

Haroenv: yep, the warning should only be in the constructor

src/utils/logger.js
eunjae-lee: https://github.com/algolia/algoliasearch-helper-js/pull/802/commits/a1c3d3deb520d15046258761a64d28be93ff85da

test/spec/algoliasearch.helper/setQueryParameter.js
eunjae-lee: https://github.com/algolia/algoliasearch-helper-js/pull/802/commits/c97cd22f9e3d43e3def417476084de50db91f941

I've put test cases in the all entries where userToken can be passed, to assure the warning message shows only once. What do you think? Too much?

src/algoliasearch.helper.js
eunjae-lee: I've talked to the search api team and confirmed that we can leave the invalid token for now.

yannickcr (approved):

Haroenv (approved):

PR #804, opened 2020-12-14
feat(answers): add `findAnswers`

eunjae-lee:
This PR adds `findAnswers` method.
In this way, we can apply facets to /answers call.

It doesn't emit `'result'` but returns a promise of result.
If we can replace the traditional hits with the hits from answers completely, then we should go for emitting `'result'`, but it's not the case at the moment. So I chose to return a promise.

Let me know what you think.

(I haven't added any tests because we haven't agreed on how to implement this in helper yet. Once decided here, I will add test.)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -172,6 +172,28 @@ declare namespace algoliasearchHelper {
     ): void;
 
     /**
+     * Start the search for answers with the parameters set in the state.
+     * This method returns a promise.
+     * @param {Object} options - the options for answers API call
+     * @param {string[]} options.attributesForPrediction - Attributes to use for predictions. If empty, `searchableAttributes` is used instead.
+     * @param {string[]} options.queryLanguages - The languages in the query. Currently only supports ['en'].
+     * @param {number} options.nbHits - Maximum number of answers to retrieve from the Answers Engine. Cannot be greater than 1000.
+     */
+    findAnswers(options: {
+      attributesForPrediction: string[];
+      queryLanguages: string[];
+      nbHits: number;
+    }): Promise<{
+      hits: Array<any & {
+        _answer: {
+          extract: string;
+          extractAttribute: string;
+          score: number;
+        }
+      }>
+    }>;
+
+    /**
      * Search for facet values based on an query and the name of a faceted attribute. This
      * triggers a search and will return a promise. On top of using the query, it also sends
      * the parameters from the state so that the search is narrowed down to only the possible values.
--- a/index.d.ts
+++ b/index.d.ts
@@ -184,7 +184,7 @@ declare namespace algoliasearchHelper {
       attributesForPrediction: string[];
       queryLanguages: string[];
       nbHits: number;
-    }): FindAnswersResponse<TObject>;
+    }): Promise<FindAnswersResponse<TObject>>;
 
     /**
      * Search for facet values based on an query and the name of a faceted attribute. This
--- a/index.d.ts
+++ b/index.d.ts
@@ -16,6 +16,7 @@ import algoliasearch, {
 import {
   SearchOptions as SearchOptionsV4,
   SearchResponse as SearchResponseV4,
+  FindAnswersResponse
   // @ts-ignore
 } from '@algolia/client-search';
 import { EventEmitter } from 'events';
@@ -179,19 +180,11 @@ declare namespace algoliasearchHelper {
      * @param {string[]} options.queryLanguages - The languages in the query. Currently only supports ['en'].
      * @param {number} options.nbHits - Maximum number of answers to retrieve from the Answers Engine. Cannot be greater than 1000.
      */
-    findAnswers(options: {
+    findAnswers<TObject>(options: {
       attributesForPrediction: string[];
       queryLanguages: string[];
       nbHits: number;
-    }): Promise<{
-      hits: Array<any & {
-        _answer: {
-          extract: string;
-          extractAttribute: string;
-          score: number;
-        }
-      }>
-    }>;
+    }): FindAnswersResponse<TObject>;
 
     /**
      * Search for facet values based on an query and the name of a faceted attribute. This
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -253,10 +253,15 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
 /**
  * Start the search for answers with the parameters set in the state.
  * This method returns a promise.
+ * @param {Object} options - the options for answers API call
+ * @param {string[]} options.attributesForPrediction - Attributes to use for predictions. If empty, `searchableAttributes` is used instead.
+ * @param {string[]} options.queryLanguages - The languages in the query. Currently only supports ['en'].
+ * @param {number} options.nbHits - Maximum number of answers to retrieve from the Answers Engine. Cannot be greater than 1000.
+ *
  * @return {AlgoliaSearchHelper}
  * @return {promise.<FacetSearchResult>} the answer results
  */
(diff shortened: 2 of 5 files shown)


src/algoliasearch.helper.js
```diff
@@ -248,6 +250,55 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
   });
 };
 
+/**
+ * Start the search for answers with the parameters set in the state.
+ * This method returns a promise.
+ * @return {AlgoliaSearchHelper}
+ * @return {promise.<FacetSearchResult>} the answer results
+ */
+AlgoliaSearchHelper.prototype.searchForAnswers = function(attributesForPrediction, queryLanguages, nbHits) {
```
Haroenv, line 259: these arguments will be more flexible in an object I think

Haroenv, line 259: the method needs to be added to the type definition still

src/algoliasearch.helper.js
```diff
@@ -248,6 +250,55 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
...
+      'hitsPerPage',
+      'restrictSearchableAttributes',
+      'clickAnalytics',
+      'facets',
+      'highlightPostTag',
+      'highlightPreTag',
+      'maxValuesPerFacet',
+      'page',
+      'snippetEllipsisText',
+      'tagFilters'
```
Haroenv, line 282: what's the reason for omitting these parameters?

eunjae-lee: It was my bad putting things in a wrong object.

src/algoliasearch.helper.js
```diff
@@ -248,6 +250,50 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
...
+    {
+      attributesForPrediction: attributesForPrediction,
+      nbHits: nbHits
+    },
+    {
+      params: omit(requestBuilder._getHitsSearchParams(derivedState), [
+        'attributesToSnippet',
+        'hitsPerPage',
+        'restrictSearchableAttributes',
+        'snippetEllipsisText' // FIXME remove this line once the engine is fixed.
```
eunjae-lee, line 276: Unlike [the doc](https://www.algolia.com/doc/rest-api/answers/#method-param-params), `snippetEllipsisText` is throwing an error. I've reported it to the team, but for now let's have it here.

src/algoliasearch.helper.js
eunjae-lee: https://github.com/algolia/algoliasearch-helper-js/pull/804/commits/e1c783f9ce843d3279b12e8ce256e6f2794073ce

eunjae-lee: https://github.com/algolia/algoliasearch-helper-js/pull/804/commits/de96b4a734e2d6d5f7c0fb3fcb4d84debf7cb350
not sure how far I should go with the typing.

index.d.ts
```diff
@@ -171,6 +171,28 @@ declare namespace algoliasearchHelper {
...
+     * @param {string[]} options.attributesForPrediction - Attributes to use for predictions. If empty, `searchableAttributes` is used instead.
+     * @param {string[]} options.queryLanguages - The languages in the query. Currently only supports ['en'].
+     * @param {number} options.nbHits - Maximum number of answers to retrieve from the Answers Engine. Cannot be greater than 1000.
+     */
+    findAnswers(options: {
+      attributesForPrediction: string[];
+      queryLanguages: string[];
+      nbHits: number;
+    }): Promise<{
+      hits: Array<any & {
```
Haroenv, line 187: can you not use the return type of findAnswers on the client?

Haroenv: this method likely should be generic, like the client one

src/algoliasearch.helper.js
```diff
@@ -248,6 +250,75 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
...
+ * @type {object}
+ * @property {Answer} _answer the object describing why the hit was chosen
+ */
+
+/**
+ * @typedef AnswersResult
+ * @type {object}
+ * @property {AnswerHit[]} hits the answer hits
+ */
+
```
Haroenv, line 272: these docs aren't used as explanation while there's the definition file (only locally), so I don't think it's needed.

Are you allowed to omit the return type all together or does eslint not allow it?

src/algoliasearch.helper.js
```diff
@@ -248,6 +250,75 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
...
+ /**
+ * Start the search for answers with the parameters set in the state.
+ * This method returns a promise.
+ * @param {Object} options - the options for answers API call
+ * @param {string[]} options.attributesForPrediction - Attributes to use for predictions. If empty, `searchableAttributes` is used instead.
+ * @param {string[]} options.queryLanguages - The languages in the query. Currently only supports ['en'].
+ * @param {number} options.nbHits - Maximum number of answers to retrieve from the Answers Engine. Cannot be greater than 1000.
+ *
+ * @return {AlgoliaSearchHelper}
```
Haroenv, line 281:
```suggestion
```

this line isn't right, you can't chain for now

src/algoliasearch.helper.js
```diff
@@ -248,6 +250,75 @@ AlgoliaSearchHelper.prototype.searchOnce = function(options, cb) {
...
+    typeof this.client.initIndex !== 'function'
+  ) {
+    throw new Error(
+      'search for answers was called, but this client does not have a function client.initIndex'
+    );
+  }
+  var index = this.client.initIndex(derivedState.index);
+  if (typeof index.findAnswers !== 'function') {
+    throw new Error(
+      'search for answers was called, but this client does not have a function client.initIndex(index).findAnswers'
```
Haroenv, line 316: this is helpful, but the second error message gives pretty much the same information. Could it be consolidated to save space? How do we do it in searchForFacetValues?

eunjae-lee: Unlike SFFV, the existence of `findAnswers` can be checked only after creating `index` here. So I couldn't merge them in a single conditional block.
https://github.com/algolia/algoliasearch-helper-js/pull/804/commits/b6d1f57923d32e198265f3dd6cdfdbc75f0ec387

index.d.ts
eunjae-lee: I've updated the dependency and now it's using the type from the client:
https://github.com/algolia/algoliasearch-helper-js/pull/804/commits/da8af1338fe8038ef29b3786be0a0c789d6a2daf

src/algoliasearch.helper.js
eunjae-lee: What do you think? https://github.com/algolia/algoliasearch-helper-js/pull/804/commits/5aa1206d303e8db790cf4fb6efa4070c3664a82d

index.d.ts
```diff
@@ -171,6 +172,20 @@ declare namespace algoliasearchHelper {
...
+     * @param {Object} options - the options for answers API call
+     * @param {string[]} options.attributesForPrediction - Attributes to use for predictions. If empty, `searchableAttributes` is used instead.
+     * @param {string[]} options.queryLanguages - The languages in the query. Currently only supports ['en'].
+     * @param {number} options.nbHits - Maximum number of answers to retrieve from the Answers Engine. Cannot be greater than 1000.
+     */
+    findAnswers<TObject>(options: {
+      attributesForPrediction: string[];
+      queryLanguages: string[];
+      nbHits: number;
+    }): FindAnswersResponse<TObject>;
```
Haroenv, line 187:
```suggestion
    }): Promise<FindAnswersResponse<TObject>>;
```

Haroenv (approved):
LGTM!

PR #809, opened 2021-02-09
fix(ts): add types for smart sort

eunjae-lee:
## Summary

This PR adds the following types for smart sort

- `relevancyStrictness` in the search parameter
- `nbSortedHits` and `appliedRelevancyStrictness` in the search response

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -511,6 +511,16 @@ declare namespace algoliasearchHelper {
     ruleContexts?: string[];
     optionalFilters?: Array<string | string[]>;
     queryLanguages?: string[];
+
+    /**
+     * A number between 0 and 100 indicating
+     * how strict the hits will be sorted.
+     * (dynamically at the query time, not in the indexing time)
+     * 
+     * 0: No smart sort at all, falling back to the original sort.
+     * 100: Very smart sort, it will provide only relevant hits.
+     */
+    relevancyStrictness?: number;
   }
 
   export class SearchParameters implements PlainSearchParameters {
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -1156,6 +1166,14 @@ declare namespace algoliasearchHelper {
      * total number of hits of this query on the index
      */
     nbHits: number;
+    /**
+     * total number of hits from smart sort
+     */
+    nbSortedHits?: number;
```
shortcuts, line 1172: We should state the word `smart` to avoid confusion with any kind of other sort, like `nbSmartSortedHits`, wdyt?

eunjae-lee: That's a property sent by the engine, so we cannot change it :)

index.d.ts
```diff
@@ -1156,6 +1166,14 @@ declare namespace algoliasearchHelper {
      * total number of hits of this query on the index
      */
     nbHits: number;
+    /**
+     * total number of hits from smart sort
```
Haroenv, line 1170: can we use the same descriptions as in the client? https://github.com/algolia/algoliasearch-client-javascript/pull/1251

eunjae-lee: Oh nice, I didn't know they were already there. https://github.com/algolia/algoliasearch-helper-js/pull/809/commits/107374a11b059135098734586cef082742bd256c

index.d.ts
```diff
@@ -511,6 +511,13 @@ declare namespace algoliasearchHelper {
     ruleContexts?: string[];
     optionalFilters?: Array<string | string[]>;
     queryLanguages?: string[];
+
+    /**
+     * The relevancy threshold to apply to search in a virtual index [0-100]. A Bigger
+     * value means fewer, but more relevant results, smaller value means more, but
+     * less relevant results.
+     */
+    relevancyStrictness?: number;
```
Haroenv, line 520: this one likely isn't needed

eunjae-lee: Can you tell me why it's not needed?
When we need to set `relevancyStrictness` from the frontend side, we will need it, no?

Haroenv: because the interface inherits from the client type, but in this case we still want to copy it for v3 compact

Haroenv (approved):

shortcuts (approved):

index.d.ts
eunjae-lee: Ah got what you mean 👍🏼

PR #811, opened 2021-02-12
fix(ts): correct type for HierarchicalFacet parameter

Haroenv:
I've checked, and the defaulting of " > " actually happens only in `_getHierarchicalFacetSeparator`, so that means you can get facets without separator. Examples:

https://github.com/algolia/instantsearch.js/blob/1ede1ae392d3a12f5b0fe29075ffeb05e572a874/src/connectors/menu/connectMenu.js#L283-L286
https://github.com/algolia/instantsearch.js/blob/1ede1ae392d3a12f5b0fe29075ffeb05e572a874/src/connectors/menu/__tests__/connectMenu-test.js#L98-L101


eunjae-lee (approved):
Makes sense

PR #822, opened 2021-06-09
feat(getFacetValues): process facetOrdering

Haroenv:
This adds support for reading from facetOrdering in getFacetValues. It's applied by default if no sortBy is given, but can be forced (or disabled) by `facetOrdering` in opts.

```js
results.getFacetValues('brand'); // facet ordering
results.getFacetValues('brand', { }); // facet ordering
results.getFacetValues('brand', { sortBy: ['isRefined:asc'] }); // no facet ordering
results.getFacetValues('brand', { sortBy: ['isRefined:asc'], facetOrdering: false }); // no facet ordering
results.getFacetValues('brand', { sortBy: ['isRefined:asc'], facetOrdering: true }); // facet ordering
```

DX-2075


src/SearchResults/index.js
```diff
@@ -659,20 +659,30 @@ function extractNormalizedFacetValues(results, attribute) {
...
 
   var children = node.data.map(function(childNode) {
-    return recSort(sortFn, childNode);
+    return recSort(sortFn, childNode, names, level + 1);
   });
-  var sortedChildren = sortFn(children);
-  var newNode = merge({}, node, {data: sortedChildren});
+  var sortedChildren = sortFn(children, names[level]);
+  var newNode = merge({}, node);
+  newNode.data = sortedChildren;
```
Haroenv, line 685: this change is because merge would otherwise cause hidden items to reappear

test/spec/SearchResults/getFacetValues-facetOrdering.js
```diff
@@ -0,0 +1,1518 @@
...
+    };
+    var results = data.content.results.slice();
+    results[0] = Object.assign(order, results[0]);
+
+    var searchParams = new SearchParameters(data.state);
+    var result = new SearchResults(searchParams, results);
+
+    var facetValues = result.getFacetValues('hierarchicalCategories');
+
+    var expected = {
```
shortcuts, line 1124: Even though we could just collapse it, I feel like exporting those variables to a file containing all of them (maybe something like `hierarchical-cards-facetOrdering-results`) could help with the readability.

Haroenv: the results are different per test though, so that wouldn't really be a good solution

shortcuts: Yes, thinking again about it it would be a complicated-naming JS file with only variables.

shortcuts (approved):
Looks good to me!

src/SearchResults/index.js
```diff
@@ -659,20 +659,30 @@ function extractNormalizedFacetValues(results, attribute) {
 }
 
 /**
- * Sort nodes of a hierarchical facet results
+ * Sort nodes of a hierarchical or disjunctive facet results
  * @private
- * @param {HierarchicalFacet} node node to upon which we want to apply the sort
+ * @param {function} sortFn
+ * @param {HierarchicalFacet|Array} node node to upon which we want to apply the sort
```
francoischalifour, line 665:
```suggestion
 * @param {HierarchicalFacet|Array} node node upon which we want to apply the sort
```

src/SearchResults/index.js
```diff
@@ -682,6 +692,70 @@ function vanillaSortFn(order, data) {
...
+
+  return orderedFacets.concat(
+    orderBy(remainingFacets, ordering[0], ordering[1])
+  );
+}
+
+/**
+ * @param {SearchResults} results the search results class
+ * @param {string} attribute the attribute to retrieve ordering of
+ * @returns {FacetOrdering=}
```
francoischalifour, line 748:
```suggestion
 * @returns {FacetOrdering}
```

src/SearchResults/index.js
francoischalifour: I'm not sure to understand. That's worth a comment imo.

src/SearchResults/index.js
```diff
@@ -682,6 +692,70 @@ function vanillaSortFn(order, data) {
...
+
+  var sortRemainingBy = facetOrdering.sortRemainingBy;
+  var ordering;
+  if (sortRemainingBy === 'hidden') {
+    return orderedFacets;
+  } else if (sortRemainingBy === 'alpha') {
+    ordering = [['name'], ['asc']];
+  } else {
+    ordering = [['count'], ['desc']];
+  }
```
francoischalifour, line 738: Making sure: we decide to fallback to sorting by count and not throwing if the provided order is wrong?

src/SearchResults/index.js
```diff
@@ -694,6 +768,9 @@ function vanillaSortFn(order, data) {
  * might not be respected if you have facet values that are already refined.
  * @param {string} attribute attribute name
  * @param {object} opts configuration options.
+ * @param {boolean} [opts.facetOrdering=false]
+ * Force the use of facetOrdering from the result if a sortBy is present. If
+ * sortBy isn't present, facetOrdering will be used automatically.
```
francoischalifour, line 773: Then it's not `false` by default? (since it's computed based on `sortBy`)

```suggestion
 * @param {boolean} [opts.facetOrdering]
 * Force the use of facetOrdering from the result if a sortBy is present. If
 * sortBy isn't present, facetOrdering will be used automatically.
```

src/SearchResults/index.js
```diff
@@ -733,30 +810,41 @@ SearchResults.prototype.getFacetValues = function(attribute, opts) {
     return undefined;
   }
 
-  var options = defaultsPure({}, opts, {sortBy: SearchResults.DEFAULT_SORT});
+  var options = defaultsPure({}, opts, {
+    sortBy: SearchResults.DEFAULT_SORT,
+    // if no sortBy is given, attempt to sort based on facetOrdering
+    // if it is given, we still allow to sort via facet ordering first
+    facetOrdering: !(opts && opts.sortBy)
```
francoischalifour, line 817: Doesn't this line override the provided option `{ facetOrdering: true }`? (Flow is hard to understand in the helper so I prefer asking.)

test/spec/SearchResults/getFacetValues-facetOrdering.js
```diff
@@ -0,0 +1,1518 @@
+'use strict';
+
+var SearchResults = require('../../../src/SearchResults');
+var SearchParameters = require('../../../src/SearchParameters');
+
+describe('disjunctive facet', function() {
+  test.each([
+    [
+      'nothing ordered',
```
francoischalifour, line 9: Tests and logic are already complex so I think we should add the expectations in the test names:

```suggestion
      'sorts by count by default',
```

Wdyt?

test/spec/SearchResults/getFacetValues-facetOrdering.js
```diff
@@ -0,0 +1,1518 @@
...
+    var results = data.content.results.slice();
+    results[0] = Object.assign(order, results[0]);
+
+    var searchParams = new SearchParameters(data.state);
+    var result = new SearchResults(searchParams, results);
+
+    var facetValues = result.getFacetValues('brand', {
+      sortBy: ['name:desc'],
+      facetOrdering: true
+    });
```
francoischalifour, line 167: That behavior is getting confusing. Shouldn't we throw instead?

test/spec/SearchResults/getFacetValues-facetOrdering.js
```diff
@@ -0,0 +1,1518 @@
...
+
+    var facetValues = result.getFacetValues('hierarchicalCategories');
+
+    var expected = {
+      name: 'hierarchicalCategories',
+      count: null,
+      isRefined: true,
+      path: null,
+      exhaustive: true,
+      data: [
```
francoischalifour, line 255: That's a lot of values, so it's easy to loose track of the expectation. Any chance we can reduce this dataset to its minimal?

francoischalifour (commented):
Great feature!

src/SearchResults/index.js
Haroenv: the equals sign means it's optional

src/SearchResults/index.js
Haroenv: it overrides right-to-left, so this is like setting defaults. Not sure why, but this option is like the reverse of object.assign

test/spec/SearchResults/getFacetValues-facetOrdering.js
Haroenv: this is needed for the hierarchical menu and menu widgets, they have a sortBy which isn't the default one, but still need to take facet ordering in account.

test/spec/SearchResults/getFacetValues-facetOrdering.js
Haroenv: yes good point, I'll copy the dataset and remove some

src/SearchResults/index.js
francoischalifour: Ah right!

src/SearchResults/index.js
francoischalifour: Right, I thought that one left to right.

test/spec/SearchResults/getFacetValues-facetOrdering.js
francoischalifour: IIUC that's an internal constraint but that constraint leaks in the public API, which is annoying—but I get the point.

Haroenv: I can see how others might want both a non-default sortBy, but still try to use facetOrdering if it's given

src/SearchResults/index.js
Haroenv: it's something that's a bit blurry at the moment, the spec mentions falling back to sortFacetValuesBy, but I can't retrieve that in the result-side, so I think falling back to just count by default is reasonable.

Alternatively we could make sortRemainingBy required in the dashboard possibly. What did you pick @VladislavFitz?

test/spec/SearchResults/getFacetValues-facetOrdering.js
Haroenv: simplified the data

src/SearchResults/index.js
Haroenv: used defaultsPure which has the expected behaviour without hack

src/SearchResults/index.js
Haroenv: this is about the key sortRemainingBy, what do you sort by if the key isn't given?

VladislavFitz: Sort by count by default sounds good to me 👍

francoischalifour (approved):

PR #823, opened 2021-06-09
feat(ts): document renderingContent

Haroenv:
Note: while the format will not change, it's not yet live

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1235,6 +1235,45 @@ declare namespace algoliasearchHelper {
     userData: any[];
 
     /**
+     * Content defining how the search interface should be rendered.
+     * This is set via the settings for a default value and can be overridden via rules
+     */
+    renderingContent?: {
+      /**
+       * defining how facets should be ordered
+       */
+      facetOrdering?: {
+        /**
+         * the ordering of facets (widgets)
+         */
+        facet?: {
+          /**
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1234,6 +1234,45 @@ declare namespace algoliasearchHelper {
...
+           */
+          order: string[];
+        };
+        /**
+         * the ordering of facet values, within an individual list
+         */
+        values?: {
+          [facet: string]: {
+            /**
+             * pinned order of facet values
```
shortcuts, line 1261:
```suggestion
             * Ordered facet values
```

index.d.ts
```diff
@@ -1234,6 +1234,45 @@ declare namespace algoliasearchHelper {
...
+      /**
+       * defining how facets should be ordered
+       */
+      facetOrdering?: {
+        /**
+         * the ordering of facets (widgets)
+         */
+        facet?: {
+          /**
+           * pinned order of facet lists
```
shortcuts, line 1251:
```suggestion
           * Ordered facet lists
```

shortcuts (approved):
It looks consistent with what is describe in the RFC and the implementation!

(My suggestions is only if we change the naming in https://github.com/algolia/algoliasearch-helper-js/pull/822#discussion_r648977242)

PR #874, opened 2021-10-06
feat(facets): when * is present, only send that parameter

Haroenv:
When `*` is present in facets, we don't need to request the individual facet values of each facet that's declared, as `*` will already request it.

We keep only `*` if it is present. If no `*` is present, the facets stay as-is.

see: https://github.com/algolia/instantsearch/issues/5723

Alternatively to this PR, similar logic could be part of the algoliasearch client cache, but that has as downside that it requires an update to the search client, also custom implementations and older clients

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -60,7 +60,7 @@ var requestBuilder = {
     var numericFilters = requestBuilder._getNumericFilters(state);
     var tagFilters = requestBuilder._getTagFilters(state);
     var additionalParams = {
-      facets: facets.indexOf('*') > 0 ? ['*'] : facets,
+      facets: facets.indexOf('*') > -1 ? ['*'] : facets,
       tagFilters: tagFilters
     };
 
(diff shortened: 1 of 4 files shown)


Haroenv:
Upon reflection, I realised that the behaviour with `-` isn't needed, that's for attributesToRetrieve, not facets

src/requestBuilder.js
```diff
@@ -60,7 +60,9 @@ var requestBuilder = {
     var numericFilters = requestBuilder._getNumericFilters(state);
     var tagFilters = requestBuilder._getTagFilters(state);
     var additionalParams = {
-      facets: facets,
+      facets: state.expandWildcardFacets
+        ? facets.filter((facet) => facet === '*' || facet[0] === '-')
```
tkrugg, line 64: why is `-` (negative filters) taken into account here. It's not currently supported is it? I don't understand why this PR tries to solve that.

Haroenv: is misread the [attributesToRetrieve](https://www.algolia.com/doc/api-reference/api-parameters/attributesToRetrieve/) docs as the facets doc

Haroenv: maybe should the parameter just be "wildcard"?

francoischalifour (approved):

PR #885, opened 2022-01-19
changed angular.js example link

cmgchess:
changed angular.js example link to algolia's codepen demo. currently it navigates to deleted directory


Haroenv (approved):
this is a good catch, thanks!

PR #888, opened 2022-02-11
fix(requestBuilder): handle filters with minus sign prefix

Jerska:
**Summary**

Fix `requestBuilder._getFacetFilters` to escape the filter value if it starts with a minus sign.


Haroenv:
I think some people might rely on this behaviour, eg. https://github.com/algolia/instantsearch.js/discussions/4754

Haroenv:
I totally understand that, but this is behaviour that has been consistent for six years, and thus can no longer be a bug fix, but would be a breaking change

Haroenv:
This has been superseded by #889, in a non-breaking manner. Thanks for your PR!

PR #889, opened 2022-02-16
feat(facetValues): offer escaped value

Haroenv:
That will be used in InstantSearch for the value.

See also #888 for another approach, where all negative refinements are automatically escaped, which is possibly a breaking change.

Still to do:
- [ ] test whether the escape is sufficient
- [ ] hierarchical values

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -619,6 +619,16 @@ SearchResults.prototype.getFacetByName = function(name) {
 };
 
 /**
+ * Replaces a leading - with \-
+ * @private
+ * @param {string} value the facet value to replace
+ * @returns string
+ */
+function escapeFacetValue(value) {
+  return value.replace(/^-/, '\\-');
+}
+
+/**
  * Get the facet values of a specified attribute from a SearchResults object.
  * @private
  * @param {SearchResults} results the search results to search in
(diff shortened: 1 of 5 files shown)


src/SearchResults/index.js
```diff
@@ -7,6 +7,7 @@ var compact = require('../functions/compact');
 var find = require('../functions/find');
 var findIndex = require('../functions/findIndex');
 var formatSort = require('../functions/formatSort');
+const escapeFacetValue = require('../functions/escapeFacetValue');
```
francoischalifour, line 10:
```suggestion
var escapeFacetValue = require('../functions/escapeFacetValue');
```

src/SearchResults/generate-hierarchical-tree.js
```diff
@@ -5,6 +5,7 @@ module.exports = generateTrees;
 var orderBy = require('../functions/orderBy');
 var find = require('../functions/find');
 var prepareHierarchicalFacetSortBy = require('../functions/formatSort');
+const escapeFacetValue = require('../functions/escapeFacetValue');
```
francoischalifour, line 8:
```suggestion
var escapeFacetValue = require('../functions/escapeFacetValue');
```

francoischalifour (approved):

src/SearchResults/index.js
Haroenv: oops, due to auto-import

dhayab (approved):

PR #890, opened 2022-03-04
fix(types): allow "__isArtificial"

Haroenv:
see https://github.com/algolia/react-instantsearch/pull/3378 for its use. As in every key of rawResults is added to the response, this does not require any code changes. You create artificial results like this: `new SearchResults(new SearchParameters(), [{ __isArtificial: true }])`

`__isArtificial` is a marker which can be added to search results to identify them as created without a search response. This is for internal use, eg. avoiding caching in infinite hits, or delaying the display of these results.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1300,7 +1300,7 @@ declare namespace algoliasearchHelper {
 
     /**
      * Marker which can be added to search results to identify them as created without a search response.
-     * This is for internal use, eg. avoiding caching in infinite hits, or delaying the display of these results.
+     * This is for internal use, e.g., avoiding caching in infinite hits, or delaying the display of these results.
      */
     __isArtificial?: boolean;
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1298,6 +1298,12 @@ declare namespace algoliasearchHelper {
     _rawResults: SearchResponse<T>[];
     _state: SearchParameters;
 
+    /**
+     * Marker which can be added to search results to identify them as created without a search response.
+     * This is for internal use, eg. avoiding caching in infinite hits, or delaying the display of these results.
```
sarahdayan, line 1303:
```suggestion
     * This is for internal use, e.g., avoiding caching in infinite hits, or delaying the display of these results.
```

sarahdayan (approved):
After discussing with @Haroenv I can see why this is a safe bet for now. We should still refactor later on so we don't have to introduce internal flags that could be hard to understand later.

Suggestion (can be tackled later): to discourage users from picking internal values, should we prefix them with something that would make it stick out in a codebase/code review, e.g., `IS_INTERNAL_ONLY_DO_NOT_USE__isArtifical`?

Haroenv:
I used `__` prefix, as `_` prefix is reserved for engine response keys, and `__` thus for frontend response keys

PR #900, opened 2022-03-31
feat(sffv): escape value for use in refinements

Haroenv:
Similar to #889, but for search for facet values.

This makes sure that the facet values match between getFacetValues and searchForFacetValues, so that isRefined works as expected too.


dhayab (approved):

src/algoliasearch.helper.js
```diff
@@ -361,6 +362,7 @@ AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxF
     content = Array.isArray(content) ? content[0] : content;
 
     content.facetHits.forEach(function(f) {
+      f.value = escapeFacetValue(f.value);
```
Haroenv, line 365: do we need the unescaped value for anything?

dhayab (approved):

PR #902, opened 2022-04-05
fix(disjunctiveFacets): avoid escaping non-string values

Haroenv:
This wasn't fully functioning behaviour before #889 already, as numeric values don't show as refined, even if they are (see https://github.com/algolia/algoliasearch-helper-js/blob/dd375ab18513336817bd8d5d78341ac33ae94954/src/SearchParameters/RefinementList.js#L141 where it's purposely casted to a string, but the input values aren't), but it didn't throw an error, which became the case in #889

A customer discovered this in https://algolia.atlassian.net/browse/CR-883

### DIFF
--- a/test/spec/SearchResults/getFacetValues.js
+++ b/test/spec/SearchResults/getFacetValues.js
@@ -237,6 +237,36 @@ test('getFacetValues(disjunctive) returns escaped facet values', function() {
   expect(facetValues.length).toBe(3);
 });
 
+test('getFacetValues introduces numeric disjunctive refinements', function() {
+  var searchParams = new SearchParameters({
+    index: 'instant_search',
+    disjunctiveFacets: ['type'],
+    disjunctiveFacetsRefinements: {
+      type: ['5', 50]
+    }
+  });
+
+  var result = {
+    query: '',
+    facets: {
+      type: {}
+    }
+  };
+
+  var results = new SearchResults(searchParams, [result, result]);
+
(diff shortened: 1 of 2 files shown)


test/spec/SearchResults/getFacetValues.js
```diff
@@ -237,6 +237,36 @@ test('getFacetValues(disjunctive) returns escaped facet values', function() {
...
+  };
+
+  var results = new SearchResults(searchParams, [result, result]);
+
+  var facetValues = results.getFacetValues('type');
+
+  var expected = [
+    {name: '5', escapedValue: '5', count: 0, isRefined: true},
+    // even though this could be considered refined, it's not because it's a number (existing bug)
+    {name: '50', escapedValue: '50', count: 0, isRefined: false}
```
Haroenv, line 263: I haven't changed the behaviour of "injected numeric facets"' isRefined, it really should be true, but that was the case before #889 already

other relevant part of the code:

https://github.com/algolia/algoliasearch-helper-js/blob/4bae51bdefd217f3ef9ab9bf02fb653f3b8ae643/src/SearchResults/index.js#L510-L519

dhayab (approved):
Should we update the unescape util in https://github.com/algolia/react-instantsearch/pull/3420 as well, or are the values cast into strings at that time?

Haroenv:
3420 is about labels, thus not relevant here :)

PR #903, opened 2022-04-08
fix(types): correct type for addTag

Haroenv:
it takes only one argument, not two

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -283,7 +283,7 @@ declare namespace algoliasearchHelper {
      * @deprecated since version 2.4.0, see {@link AlgoliaSearchHelper#addFacetExclusion}
      */
     addExclude: AlgoliaSearchHelper['addFacetExclusion'];
-    addTag(facet: string, value: string): this;
+    addTag(facet: string): this;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -283,7 +283,7 @@ declare namespace algoliasearchHelper {
      * @deprecated since version 2.4.0, see {@link AlgoliaSearchHelper#addFacetExclusion}
      */
     addExclude: AlgoliaSearchHelper['addFacetExclusion'];
-    addTag(facet: string, value: string): this;
+    addTag(facet: string): this;
```
dhayab, line 286: Nitpicky suggestion:

```suggestion
    addTag(value: string): this;
```

or, if we follow https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/algoliasearch.helper.js#L581:

```suggestion
    addTag(tag: string): this;
```

dhayab (approved):

index.d.ts
Haroenv: good call, thanks, should have removed the first, not the second part :D

PR #906, opened 2022-05-16
feat(searchForFacetValues): fall back to client.search if it's present

Haroenv:
client.search with `type: 'facet', facet, params: { facetQuery }` can be used in v4 onwards as an alternative to client.searchForFacetValues, meaning the client.searchForFacetValues method can be removed in a future major version.

Unfortunately the JS client v3 doesn't support type "facet" for multi-index search because it manually decides which keys are used in the post object: https://github.com/algolia/algoliasearch-client-javascript/blob/v3/src/AlgoliaSearchCore.js#L638-L654

FX-1432

### DIFF
--- a/src/algoliasearch.helper.js
+++ b/src/algoliasearch.helper.js
@@ -330,30 +330,57 @@ AlgoliaSearchHelper.prototype.findAnswers = function(options) {
  */
 AlgoliaSearchHelper.prototype.searchForFacetValues = function(facet, query, maxFacetHits, userState) {
   var clientHasSFFV = typeof this.client.searchForFacetValues === 'function';
+  var clientHasInitIndex = typeof this.client.initIndex === 'function';
   if (
     !clientHasSFFV &&
-    typeof this.client.initIndex !== 'function'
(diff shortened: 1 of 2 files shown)


src/algoliasearch.helper.js
```diff
@@ -330,31 +330,58 @@ AlgoliaSearchHelper.prototype.findAnswers = function(options) {
...
+    ]);
+    // algoliasearch < 3.27.1
+  } else if (clientHasInitIndex) {
+    searchForFacetValuesPromise = this.client
+      .initIndex(state.index)
+      .searchForFacetValues(algoliaQuery);
+    // algoliasearch ~5.0.0
+  } else {
+    // @MAJOR only use client.search
+    delete algoliaQuery.facetName;
```
dhayab, line 364: Why is this removed from the params payload?

Haroenv:
Marked as a draft as the signature for v5 and v4 aren't equal wrt params

dhayab (approved):

francoischalifour (approved):

PR #910, opened 2022-06-15
feat(types): support algoliasearch v5

Haroenv:
Algoliasearch v5 has different type names and packages, so needs to be supported separately.

This pull request has the following changes:

1. add a folder `/types/algoliasearch` with both a .d.ts and .js file
2. add utilities and simplify the dealing of importing different types across the major versions

FX-1434

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "@algolia/events": "^4.0.1"
   },
   "peerDependencies": {
-    "algoliasearch": ">= 3.1 < 5"
+    "algoliasearch": ">= 3.1 < 6"
   },
   "resolutions": {
     "node-sass": "4.14.1"
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "dist",
     "src",
     "index.js",
-    "index.d.ts"
+    "index.d.ts",
+    "types/algoliasearch.d.ts"
   ],
   "devDependencies": {
     "@types/algoliasearch": "3.34.11",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "src",
     "index.js",
     "index.d.ts",
-    "types/algoliasearch.d.ts"
+    "types/algoliasearch.d.ts",
+    "types/algoliasearch.js"
   ],
   "devDependencies": {
     "@types/algoliasearch": "3.34.11",
--- a/types/algoliasearch.d.ts
+++ b/types/algoliasearch.d.ts
@@ -5,8 +5,6 @@
 // @ts-ignore
 import type algoliasearch from 'algoliasearch/lite';
 // @ts-ignore
-import type * as AlgoliaSearchLite from 'algoliasearch/lite';
-// @ts-ignore
 import type * as AlgoliaSearchLiteV5 from 'algoliasearch-lite';
 // @ts-ignore
 import type * as AlgoliaSearch from 'algoliasearch';
@@ -58,7 +56,7 @@ type DefaultSearchClient = PickForClient<{
 
 export type SearchOptions = PickForClient<{
   // @ts-ignore
-  v3: AlgoliaSearchLite.QueryParameters;
+  v3: AlgoliaSearch.QueryParameters;
   // @ts-ignore
   v4: ClientSearch.SearchOptions;
   v5: NonNullable<
@@ -69,7 +67,7 @@ export type SearchOptions = PickForClient<{
 
 export type SearchResponse<T> = PickForClient<{
   // @ts-ignore
-  v3: AlgoliaSearchLite.Response<T>;
+  v3: AlgoliaSearch.Response<T>;
   // @ts-ignore
   v4: ClientSearch.SearchResponse<T>;
   // @ts-ignore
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -101,7 +101,7 @@
     "@algolia/events": "^4.0.1"
   },
   "peerDependencies": {
-    "algoliasearch": ">= 3.1 < 5"
+    "algoliasearch": ">= 3.1 < 6"
```
Haroenv, line 104: technically this is now algoliasearch or algoliasearch-lite, but I don't know how to express that. Both being optional defeats the point

types/algoliasearch.d.ts
```diff
@@ -0,0 +1,124 @@
...
+// @ts-ignore
+import type * as AlgoliaSearchLite from 'algoliasearch/lite';
+// @ts-ignore
+import type * as ClientSearch from '@algolia/client-search';
+// @ts-ignore
+import type * as AlgoliaSearchLiteV5 from '@experimental-api-clients-automation/algoliasearch-lite';
+// @ts-ignore
+import type * as AlgoliaSearchFullV5 from '@experimental-api-clients-automation/algoliasearch';
+// @ts-ignore
+import * as ClientSearchV5 from '@experimental-api-clients-automation/client-search';
```
Haroenv, line 16: these get renamed to @algolia at public beta, so will probably be able to be merged into the other imports

types/algoliasearch.d.ts
```diff
@@ -0,0 +1,148 @@
...
+            sortRemainingBy?: 'count' | 'alpha' | 'hidden';
+          };
+        };
+      };
+    };
+  };
+  // @ts-ignore
+  v4: ClientSearch.SearchResponse<T>;
+  // @ts-ignore
+  v5: AlgoliaSearch.SearchResponse; // TODO: should be generic
```
Haroenv, line 102: debating with @shortcuts whether SearchResponse already can have this from the alpha, as you can't give a generic to something that doesn't have it, or detect when something is or isn't generic

Haroenv: if not possible, we can ship like this and later add the generic when it will be generic, but that means there's a moment where v5 is supported but the response isn't generic in eg. hits

package.json
sarahdayan: For expressing optionality, you can use [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta).

Is it necessary to specify `algoliasearch/lite` anyway? It's not even its own package.

types/algoliasearch.d.ts
```diff
@@ -0,0 +1,148 @@
...
+  ? unknown
+  : T;
+
+type DummySearchClientV4 = {
+  transporter: unknown;
+};
+
+type DummySearchClient = {
+  search: unknown;
+};
```
sarahdayan, line 25: Are these `Dummy*` types already being used and accessible to users? The term is ableist so if we have the ability to rename it, we should.

types/algoliasearch.d.ts
sarahdayan: Why isn't the v5 `SearchResponse` generic for now?

types/algoliasearch.js
```diff
@@ -0,0 +1 @@
+// fake file to work with the types
```
sarahdayan, line 1: Can you elaborate a bit in this comment? It's not fully clear when reading it out of context.

package.json
shortcuts: > Is it necessary to specify algoliasearch/lite anyway? It's not even its own package.

I believe the comment refer to the change prior to this PR https://github.com/algolia/api-clients-automation/pull/729, it's not necessary anymore!

types/algoliasearch.d.ts
shortcuts: It's generated code so we did not (yet) focused on implementing it in some places (`search`/`browse` for example), but will do before stable!

sarahdayan: This means we'll need to provide a default value in the helper when it becomes generic so it doesn't break, which is fine. Do you see any possible issues @Haroenv?

Haroenv: The issue I see is that we will have to say "if you're using 3.9.0 of the helper, it's supported, if you're using 3.10.0 (or whatever version has the final) it's generic". It's not that big of a deal as you can just say to update of course

types/algoliasearch.d.ts
Haroenv: I'd not heard it in the insulting context before (only as a model). They aren't exposed so we can rename to whatever we want. is MockSearchClient or SearchClientShape fine?

types/algoliasearch.js
Haroenv:
```suggestion
// fake file to allow `export * from 'algoliasearch-helper/types/algoliasearch'`
```

is that good?

types/algoliasearch.d.ts
sarahdayan: Yeah it's listed in the ableist terms to avoid of the Algolia docs (inherited from the [Google styleguide](https://developers.google.com/style/inclusive-documentation#ableist-language)), and many other sources flag it as a problematic term.

`MockSearchClient` sounds fine, especially if that's a type we could reuse for tests, that'd make sense to me.

sarahdayan (approved):

PR #911, opened 2022-06-17
fix(requests): send a sorted object of parameters

Haroenv:
Having a sorted set of parameters helps with making sure the cache is hit in more cases.

For example, when you remove and immediately add the same widget again, and they no longer are in the same order, but have the same parameters conceptually.

This PR does not sort the items inside the parameters, eg. facets, which would be possible, but isn't done yet.

FX-1473

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -2,6 +2,17 @@
 
 var merge = require('./functions/merge');
 
+function sortObject(obj) {
+  return Object.keys(obj)
+    .sort(function(a, b) {
+      return a.localeCompare(b);
+    })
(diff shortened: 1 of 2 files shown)


src/requestBuilder.js
```diff
@@ -2,6 +2,17 @@
 
 var merge = require('./functions/merge');
 
+function sortObject(obj) {
```
Haroenv, line 5: I put it in this file instead of functions as I wasn't sure it has uses outside of here

francoischalifour (approved):
Useful for reducing network requests sent from conditional widgets and dynamic widgets!

PR #912, opened 2022-06-23
feat(disjunctiveFacetParams): reduce payload size

Jerska:
This PR reduces the amount of params sent for each disjunctive facet request.
It also reduces `hitsPerPage` from `1` to `0`, allowing us to remove the `attributes*` parameters at the same time.

This reduces:
* the payload sent by a few bytes
* and the response size by a few extra bytes as the previous implem sent back one `objectID` per additional request

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -121,8 +121,7 @@ var requestBuilder = {
       hitsPerPage: 0,
       page: 0,
       analytics: false,
-      clickAnalytics: false,
-      responseFields: ['facets', 'renderingContent']
+      clickAnalytics: false
     };
 
     if (tagFilters.length > 0) {
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -121,7 +121,8 @@ var requestBuilder = {
       hitsPerPage: 0,
       page: 0,
       analytics: false,
-      clickAnalytics: false
+      clickAnalytics: false,
+      responseFields: ["facets", "renderingContent"],
     };
     
     if (tagFilters.length > 0) {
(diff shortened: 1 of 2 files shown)


src/requestBuilder.js
```diff
@@ -118,15 +118,15 @@ var requestBuilder = {
...
     var additionalParams = {
-      hitsPerPage: 1,
+      hitsPerPage: 0,
       page: 0,
-      attributesToRetrieve: [],
-      attributesToHighlight: [],
-      attributesToSnippet: [],
-      tagFilters: tagFilters,
       analytics: false,
       clickAnalytics: false
```
Jerska, line 124:
```suggestion
      clickAnalytics: false,
      responseFields: ["facets"],
```

src/requestBuilder.js
```diff
@@ -118,15 +118,16 @@ var requestBuilder = {
...
+      hitsPerPage: 0,
       page: 0,
-      attributesToRetrieve: [],
-      attributesToHighlight: [],
-      attributesToSnippet: [],
-      tagFilters: tagFilters,
       analytics: false,
-      clickAnalytics: false
+      clickAnalytics: false,
+      responseFields: ["facets", "renderingContent"],
```
Haroenv, line 125:
```suggestion
      responseFields: ['facets', 'renderingContent']
```

src/requestBuilder.js
```diff
@@ -118,15 +118,16 @@ var requestBuilder = {
...
     };
+    
+    if (tagFilters.length > 0) {
+      additionalParams.tagFilters = tagFilters; 
+    }
```
Jerska, line 130:
```suggestion

    if (tagFilters.length > 0) {
      additionalParams.tagFilters = tagFilters; 
    }
```

Jerska:
Yes I'm on it, but I'm currently struggling with the `node-sass` version not being compatible with `node` 16 ^^"

Haroenv (approved):

sarahdayan:
We can replace `node-sass` with `dart-sass` in another PR since the former is deprecated.

PR #916, opened 2022-09-26
fix(facetValues): use additional queries to retrieve refined facet values

dhayab:
**Issue**

When a refinement list is added to an InstantSearch application, its values are requested with the hits. By default, 10 values are requested per facet.

An option available in refinement lists allows the user to search and refine on specific facet values matching the search query. This creates an issue where sometimes, the refined facet value is not returned as part of the truncated list. For example, you might refine on "Zimbabwe" but the results will contain a list of 10 facet values that won't include it, and there is no way to ask the engine to [sort facet values](https://www.algolia.com/doc/api-reference/api-parameters/sortFacetValuesBy/) by refinement state at that stage.

The result is that once refined, a facet value disappears from the list, and can only be visible again if the user search for it in a searchable refinement list widget.

[CodeSandbox →](https://codesandbox.io/s/xenodochial-bash-kr1bck?file=/src/app.js)  
_(refine on "Partner/REI")_

**Solution**

The solution I have devised adds additional queries to the search request, to get the [maximum number of facet values](https://www.algolia.com/doc/api-reference/api-parameters/maxValuesPerFacet/) allowed for each refined facet.

Then, when facet values are requested, we check that every refined facet value from the state is present in the updated list of facet values. If one isn't, its name and associated count is added from the additional query results.

This solution has a caveat though, when the refined facet value is not present in the additional query results, because the facet contains than 1000 values. In this case, we still add the facet value, but its associated count is set to `0`.

[FX-1682](https://algolia.atlassian.net/browse/FX-1682)

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -29,12 +29,24 @@ var requestBuilder = {
       params: requestBuilder._getHitsSearchParams(state)
     });
 
-    // One for each disjunctive facets
+    // One for each disjunctive facet
     state.getRefinedDisjunctiveFacets().forEach(function(refinedFacet) {
       queries.push({
         indexName: index,
         params: requestBuilder._getDisjunctiveFacetSearchParams(state, refinedFacet)
       });
+    });
+
+    // One for each conjunctive facet
+    state.facets.forEach(function(facet) {
+      if (state.getConjunctiveRefinements(facet).length > 0) {
+        var queryParams = requestBuilder._getDisjunctiveFacetSearchParams(state, facet);
+        queryParams.maxValuesPerFacet = 1000;
+        queries.push({
+          indexName: index,
+          params: queryParams
+        });
+      }
     });
 
     // More to get the parent levels of the hierarchical facets when refined
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -644,6 +644,25 @@ function extractNormalizedFacetValues(results, attribute) {
     var facet = find(results.facets, predicate);
     if (!facet) return [];
 
+    // generate a map with all the refined facet values and their actual count
+    if (!facet._facetHitsMap) {
+      facet._facetHitsMap = results._rawResults.slice(1).reduce(function(facetHitsMap, rawResult) {
+        if (rawResult.facets[attribute]) {
+          Object.assign(facetHitsMap, rawResult.facets[attribute]);
+        }
+        return facetHitsMap;
+      }, {});
+    }
+
+    // add the conjunctive refinements from the map
+    // if they have not been retrieved
+    results._state.getConjunctiveRefinements(attribute).forEach(function(refinedValue) {
+      var unescapedFacetValue = unescapeFacetValue(refinedValue);
+      if (!facet.data[unescapedFacetValue]) {
+        facet.data[unescapedFacetValue] = facet._facetHitsMap[unescapedFacetValue] || 0;
+      }
+    });
+
     return Object.keys(facet.data).map(function(name) {
       var value = escapeFacetValue(name);
       return {
--- a/test/spec/SearchResults/getFacetValues.js
+++ b/test/spec/SearchResults/getFacetValues.js
@@ -20,6 +20,29 @@ test('getFacetValues(facetName) returns a list of values using the defaults', fu
 });
 
 test(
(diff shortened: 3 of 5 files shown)


Haroenv (commented):
Is this something that should be done without option? if people use facetOrdering, the selected value doesn't get propagated to the top of the list, and thus we'd be making the second list a little bit uselessly (although it would give us a better idea of the order if it's outside of the first `showMoreLimit`).

Is 1000 the right threshold? I think it may make more sense to ask the engine to make sure a refined facet is returned inside the limit by default maybe?

I'm thinking these things because it could add some latency due to more processing and a larger payload

dhayab:
@Haroenv 

> Is this something that should be done without option? if people use facetOrdering, the selected value doesn't get propagated to the top of the list, and thus we'd be making the second list a little bit uselessly (although it would give us a better idea of the order if it's outside of the first showMoreLimit).

I don't think we know that facetOrdering is used at the stage the queries are created.

> Is 1000 the right threshold? I think it may make more sense to ask the engine to make sure a refined facet is returned inside the limit by default maybe?

That's a great idea, although it might take some time before something is implemented. We already send them the refined facet values with `facetFilters`, we could have a flag telling them to explicitly add those facet values to the list.

```js
{
    "indexName": "instant_search",
    "params": {
        "facetFilters": [
            "tags:Partner/REI"
        ],
        "facets": [
            "tags"
        ],
        "includeFacetFilters": true, // appends the values in facet filters to the retrieved facets lists
    }
}
```

Because we don't know what sort options are defined, we could have a list that have more than `maxValuesPerFacet`, so that after sorting / limit is applied, it still has the expected number and type of values.

In the meantime, we could have this fix as a temporary solution.

> I'm thinking these things because it could add some latency due to more processing and a larger payload

That's definitely a domain I'm not familiar with, I'll try to find some resources that could help me have this in mind in future developments.

src/requestBuilder.js
```diff
@@ -29,14 +29,26 @@ var requestBuilder = {
...
-    // One for each disjunctive facets
+    // One for each disjunctive facet
     state.getRefinedDisjunctiveFacets().forEach(function(refinedFacet) {
       queries.push({
         indexName: index,
         params: requestBuilder._getDisjunctiveFacetSearchParams(state, refinedFacet)
       });
     });
 
+    // One for each conjunctive facet
```
seafoox, line 40: @dhayab I'm I correct understanding 

1. that we're sending additional search request. In the case N additional request(s) where N is the number of facet(s) in the InstantSearch instance, and which have at least one active / refined facet value? 
2. that for those extra request we retrieve 1000 values, increasing at the same time the response payload?

dhayab: That's exactly that @seafoox.

Those are additional queries agregated in a single network request, thus having no pricing impact on the customer. The search parameters for those queries set `hitsPerPage: 0` because we don't need that, but they can indeed include up to 1000 facet values each.

Haroenv (changes requested):
on reflection, I don't think we should solve this issue by doing extra requests. What I have in mind is letting those customers know to put `configure({maxValuesPerFacet: 1000})` (limit will still be respected) rather than doing an extra query altogether, as we don't know the performance implications.

I'm however a fan of trying to control which facets are returned more detailed with the engine in a separate project

dhayab:
@Haroenv I think that's a good idea.

I can add an entry [in the FAQ](https://www.algolia.com/doc/guides/building-search-ui/troubleshooting/faq/js/) for all InstantSearch flavors that documents why refined facet values might not appear and how they can set `maxValuesPerFacet` to address this.

@seafoox wdyt? Is there a more suitable location for sharing this information or is the FAQ a good place?

seafoox:
@dhayab I'm worried we change that behaviour seamlessly, as it might have a direct impact on the invoice of customers on older pricing version (due to the extra call), but also to the performance overall. So sharing how to work around in the docs sounds safer to me at this stage, and we can consider changing the bahviour with the next major. 

Re on where this should leave, your proposal of adding an FAQ entry sounds good to me. Do you think in addition we should link to that FAQ entry from the reference docs https://www.algolia.com/doc/api-reference/widgets/refinement-list/js/#widget-param-sortby?

dhayab:
Good idea :) ! I'm closing this PR then, and will start working on documentation.

PR #917, opened 2022-10-03
feat(searchForFacetValues): hierarchical facet isRefined and search state clearing

cdhawke:
This PR adds the ability to mark hierarchical facets as refined, as well as clear the search refinement state on searching for a hierarchical facet. 

The motivation behind this is to add search functionality to the `Menu` instantsearch widget, which utilizes hierarchical facets for refinement. In order to be able to search through these (single level) facets, we need to mark the results as refined, and clear the search state while building the request so that the current refinements aren't taken into account during the search.

See https://github.com/algolia/instantsearch.js/tree/feat/searchable-menu for reference

[SRCH-5213]

[SRCH-5213]: https://algolia.atlassian.net/browse/SRCH-5213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -360,7 +360,9 @@ var requestBuilder = {
   },
 
   getSearchForFacetQuery: function(facetName, query, maxFacetHits, state) {
-    var stateForSearchForFacetValues = state.isDisjunctiveFacet(facetName) ?
+    // Clear the refinement state before searching for disjunctive or hierarchical facets, because we
+    // want to search the entire list of facet values, not just the subset of refined ones.
+    var stateForSearchForFacetValues = state.isDisjunctiveFacet(facetName) || state.isHierarchicalFacet(facetName) ?
       state.clearRefinements(facetName) :
       state;
     var searchForFacetSearchParameters = {
--- a/src/requestBuilder.js
+++ b/src/requestBuilder.js
@@ -359,10 +359,14 @@ var requestBuilder = {
     return hierarchicalFacet.attributes.slice(0, parentLevel + 1);
   },
 
-  getSearchForFacetQuery: function(facetName, query, maxFacetHits, state) {
+  getSearchForFacetQuery: function(facetName, query, maxFacetHits, state, includeHierarchical) {
     // Clear the refinement state before searching for disjunctive or hierarchical facets, because we
     // want to search the entire list of facet values, not just the subset of refined ones.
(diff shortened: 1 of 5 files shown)


src/requestBuilder.js
```diff
@@ -360,7 +360,9 @@ var requestBuilder = {
   },
 
   getSearchForFacetQuery: function(facetName, query, maxFacetHits, state) {
-    var stateForSearchForFacetValues = state.isDisjunctiveFacet(facetName) ?
+    // Clear the refinement state before searching for disjunctive or hierarchical facets, because we
+    // want to search the entire list of facet values, not just the subset of refined ones.
+    var stateForSearchForFacetValues = state.isDisjunctiveFacet(facetName) || state.isHierarchicalFacet(facetName) ?
```
Haroenv, line 365: I don't think we have a test for this now, but it would be interesting to ensure it stays as-is

Haroenv (commented):
I've made a playground for this: https://codesandbox.io/s/javascript-helper-app-forked-o5l78f?file=/src/app.ts

This change makes sense to me, and in theory nobody would have relied on this behaviour using InstantSearch, so there it's not a breaking change really, but with usage of just the helper, people could assume that sffv for hierarchical facets returned the refined value. Maybe it should be an option? happy to explore further though, this is a great step!

Haroenv (approved):

src/requestBuilder.js
```diff
@@ -359,8 +359,14 @@ var requestBuilder = {
     return hierarchicalFacet.attributes.slice(0, parentLevel + 1);
   },
 
-  getSearchForFacetQuery: function(facetName, query, maxFacetHits, state) {
-    var stateForSearchForFacetValues = state.isDisjunctiveFacet(facetName) ?
+  getSearchForFacetQuery: function(facetName, query, maxFacetHits, state, includeHierarchical) {
+    // Clear the refinement state before searching for disjunctive or hierarchical facets, because we
+    // want to search the entire list of facet values, not just the subset of refined ones.
+    var shouldClear = includeHierarchical
```
Haroenv, line 365:
```suggestion
    // @MAJOR: only support includeHierarchical === true
    var shouldClear = includeHierarchical
```

cdhawke:
@Haroenv im assuming we should close this given the searchable menu closure?

Haroenv:
True, this would be easily revivable, but if the complexity wouldn't be used in InstantSearch, we shouldn't add it until then. Thanks a lot for this, it was very interesting and useful!

PR #923, opened 2023-01-09
fix: prevent prototype pollution in rare error-cases

### LINKED ISSUE #922: Prototype pollution found in merge.js

Haroenv:
If a user-provided search parameter is used to instantiate search parameters, it was possible to construct it in such a way that `constructor.prototype` is attempted to be written. That throws an error, but if the error would be caught, the resulting injection still happened.

This PR fixes that (small) vulnerability by ensuring `constructor`, is skipped, just like `__proto__`.

fixes #922

This is similar/a follow-up to #880


dhayab (approved):
👍

PR #925, opened 2023-01-12
fix(getFacetValues): reflect the value of _state in hierarchicalFacetValues

Haroenv:
Before this commit, the behaviour of conjunctive/disjunctive facets and hierarchical facets is different when `results._state` has changed since the instantiation of the SearchResults class.

Conjunctive/disjunctive take the current value of `_state`, while `hierarchical` took `isRefined` from the constructor.

We want to make this change to enable optimistic UI, where we render the existing results with the next state. This causes the `isRefined` state to immediately be useful for display.

Regardless of optimistic UI, this change still makes sense to do, as it makes the behaviour of getFacetValues consistent between all types of facet.

FX-2201

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -668,7 +668,33 @@ function extractNormalizedFacetValues(results, attribute) {
       };
     });
   } else if (results._state.isHierarchicalFacet(attribute)) {
-    return find(results.hierarchicalFacets, predicate);
+    var hierarchicalFacetValues = find(results.hierarchicalFacets, predicate);
+    if (!hierarchicalFacetValues) return hierarchicalFacetValues;
+
+    var hierarchicalFacet = results._state.getHierarchicalFacetByName(attribute);
+    var currentRefinementSplit = unescapeFacetValue(
+      results._state.getHierarchicalRefinement(attribute)[0] || ''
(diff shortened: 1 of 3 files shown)


src/SearchResults/index.js
```diff
@@ -668,7 +668,33 @@ function extractNormalizedFacetValues(results, attribute) {
...
+    var hierarchicalFacetValues = find(results.hierarchicalFacets, predicate);
+    if (!hierarchicalFacetValues) return hierarchicalFacetValues;
+
+    var hierarchicalFacet = results._state.getHierarchicalFacetByName(attribute);
+    var currentRefinementSplit = unescapeFacetValue(
+      results._state.getHierarchicalRefinement(attribute)[0] || ''
+    ).split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet));
+    currentRefinementSplit.unshift(attribute);
+
+    setIsRefined(hierarchicalFacetValues, currentRefinementSplit, 0);
```
Haroenv, line 680: this option is mutating, as recreating the hierarchical tree requires information that is no longer available at this point using generateHierarchicalTree as far as I can tell

dhayab (approved):

FabienMotte (approved):

PR #938, opened 2023-04-12
feat(DerivedHelper): skip request for empty index

Haroenv:
We currently don't introspect, but if the indexName is empty, the search client will return an error, meaning skipping those queries isn't really a breaking change.

If no queries would be triggered at all, client.search is also not called.

Skipping those queries can be interesting to create patterns like "multiple sibling indices at root of instantsearch" and "orphan indices".


dhayab (approved):
Great!

PR #940, opened 2023-05-15
fix(SearchResults): use empty facets object for exclusion when results are artificial

### LINKED ISSUE #921: exclude refinement results doesn't check whether the mainSubResponse.facets exist
We call this unconditionally, meaning that when a wrong result (eg fake one from React InstantSearch Hooks) is used, this throws an error. We need to change from 

```js
  // add the excludes
  Object.keys(state.facetsExcludes).forEach(function(facetName) {
    var excludes = state.facetsExcludes[facetName];
    var position = facetsIndices[facetName];

    self.facets[position] = {
      name: facetName,
      data: mainSubResponse.facets[facetName],
      exhaustive: mainSubResponse.exhaustiveFacetsCount
    };
    excludes.forEach(function(facetValue) {
      self.facets[position] = self.facets[position] || {name: facetName};
      self.facets[position].data = self.facets[position].data || {};
      self.facets[position].data[facetValue] = 0;
    });
  });
```

to 

```js
  // add the excludes
  Object.keys(state.facetsExcludes).forEach(function(facetName) {
    var excludes = state.facetsExcludes[facetName];
    var position = facetsIndices[facetName];

    self.facets[position] = {
      name: facetName,
      data: mainFacets[facetName],
      exhaustive: mainSubResponse.exhaustiveFacetsCount
    };
    excludes.forEach(function(facetValue) {
      self.facets[position] = self.facets[position] || {name: facetName};
      self.facets[position].data = self.facets[position].data || {};
      self.facets[position].data[facetValue] = 0;
    });
  });
```

dhayab:
This PR ensures that facet exclusions are computed from a result's facets property that is defaulted to an empty object, in case this is called with artificial results.

FX-2179
Closes #921

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -588,7 +588,7 @@ function SearchResults(state, results, options) {
 
     self.facets[position] = {
       name: facetName,
-      data: mainSubResponse.facets[facetName],
+      data: mainFacets[facetName],
       exhaustive: mainSubResponse.exhaustiveFacetsCount
     };
     excludes.forEach(function(facetValue) {
--- a/test/spec/SearchResults/getRefinements.js
+++ b/test/spec/SearchResults/getRefinements.js
@@ -64,6 +64,27 @@ test('getRefinements(facetName) returns a refinement(exclude) when a facet exclu
   expect(hasSameNames(refinements, refinedFacetValues)).toBeTruthy();
 });
 
+// See: https://github.com/algolia/algoliasearch-helper-js/issues/921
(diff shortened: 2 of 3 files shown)


FabienMotte (approved):

sheldonj:
@dhayab @FabienMotte  -- how come this commit isn't in the 3.13.0 release? Feels like it should be but checking the changes here. It seems like this commit never made it to that release?

https://www.npmjs.com/package/algoliasearch-helper?activeTab=code

Haroenv:
There hasn't yet been a release since this commit, but it indeed makes sense to release as we have nothing else planned to release yet. Doing it now, thanks!

sheldonj:
Thank you very much!!

PR #946, opened 2023-06-15
test(unit): add retry mechanism

Haroenv:
The tests are sometimes flaky (integration mostly), as we're not entirely sure what the cause is, retrying seems sensible

This can also be reused in the monorepo later

### DIFF
--- a/scripts/retry.sh
+++ b/scripts/retry.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+# Retry a command up to a specific number of times until it exits successfully
+# Usage: scripts/retry.sh <retries> <command>
+
+retries="$1"
+command="$2"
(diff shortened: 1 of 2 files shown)


scripts/retry.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+# Retry a command up to a specific number of times until it exits successfully
+# Usage: scripts/retry.sh <retries> <command>
+
+retries="$1"
+command="$2"
+
+echo tyring "$command", $retries retries left
```
aymeric-giraudet, line 8:
```suggestion
echo trying "$command", $retries retries left
```

aymeric-giraudet (approved):
Cool !

PR #947, opened 2023-06-16
chore(tools): update eslint and make all rules pass

Haroenv:
A couple notable changes are:
- prettier is still disabled
- code is still in ES5/CommonJS
- jsdoc is slightly more complete
- exact same versions of eslint and plugins as in monorepo are used
- some tests are slightly cleaned up (removed duplicate and using test.each)
- no actual code should have changed (this is important, please focus the review on this)

FX-2333


dhayab (approved):
Looks good to me, conform to what you described.

PR #949, opened 2023-06-19
fix(getFacetValues): ignore `rootPath` for hierarchical facets

aymeric-giraudet:
### [FX-2414](https://algolia.atlassian.net/browse/FX-2414) and [CR-3625](https://algolia.atlassian.net/browse/CR-3625)

Just removing the `rootPath` from `currentRefinementSplit` works !

[FX-2414]: https://algolia.atlassian.net/browse/FX-2414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CR-3625]: https://algolia.atlassian.net/browse/CR-3625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -716,7 +716,11 @@ function extractNormalizedFacetValues(results, attribute) {
       results._state.getHierarchicalFacetByName(attribute);
     var currentRefinementSplit = unescapeFacetValue(
       results._state.getHierarchicalRefinement(attribute)[0] || ''
-    ).split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet));
+    )
+      .split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet))
+      .filter(function (name) {
+        return name !== hierarchicalFacet.rootPath;
+      });
     currentRefinementSplit.unshift(attribute);
 
     setIsRefined(hierarchicalFacetValues, currentRefinementSplit, 0);
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -720,7 +720,7 @@ function extractNormalizedFacetValues(results, attribute) {
       results._state.getHierarchicalRefinement(attribute)[0] || ''
     );
 
-    if (currentRefinement.startsWith(hierarchicalFacet.rootPath)) {
+    if (currentRefinement.indexOf(hierarchicalFacet.rootPath) === 0) {
       currentRefinement = currentRefinement.replace(
         hierarchicalFacet.rootPath + separator,
         ''
--- a/src/SearchResults/index.js
+++ b/src/SearchResults/index.js
@@ -714,13 +714,20 @@ function extractNormalizedFacetValues(results, attribute) {
 
     var hierarchicalFacet =
       results._state.getHierarchicalFacetByName(attribute);
-    var currentRefinementSplit = unescapeFacetValue(
+    var separator =
+      results._state._getHierarchicalFacetSeparator(hierarchicalFacet);
+    var currentRefinement = unescapeFacetValue(
       results._state.getHierarchicalRefinement(attribute)[0] || ''
-    )
-      .split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet))
-      .filter(function (name) {
-        return name !== hierarchicalFacet.rootPath;
-      });
+    );
+
+    if (hierarchicalFacet.rootPath) {
+      currentRefinement = currentRefinement.replace(
+        hierarchicalFacet.rootPath + separator,
+        ''
+      );
+    }
+
+    var currentRefinementSplit = currentRefinement.split(separator);
     currentRefinementSplit.unshift(attribute);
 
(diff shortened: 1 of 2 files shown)


src/SearchResults/index.js
```diff
@@ -716,7 +716,7 @@ function extractNormalizedFacetValues(results, attribute) {
       results._state.getHierarchicalFacetByName(attribute);
     var currentRefinementSplit = unescapeFacetValue(
       results._state.getHierarchicalRefinement(attribute)[0] || ''
-    ).split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet));
+    ).split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet)).filter(name => name !== hierarchicalFacet.rootPath);
```
Haroenv, line 719:
```suggestion
    ).split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet)).filter(function(name) { return name !== hierarchicalFacet.rootPath});
```

(and formatted)

src/SearchResults/index.js
```diff
@@ -716,7 +716,11 @@ function extractNormalizedFacetValues(results, attribute) {
       results._state.getHierarchicalFacetByName(attribute);
     var currentRefinementSplit = unescapeFacetValue(
       results._state.getHierarchicalRefinement(attribute)[0] || ''
-    ).split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet));
+    )
+      .split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet))
+      .filter(function (name) {
+        return name !== hierarchicalFacet.rootPath;
+      });
```
Haroenv, line 723: Maybe it already works, but not sure from just looking at this + the test, but:

- what if rootPath is multiple segments long, how does this comparison work? (I guess name is the path so it works)
- What if rootPath is repeated later in the path, like "root > something > root", will root be removed twice? Maybe we should return once it's been removed once or similar?

aymeric-giraudet: [The docs say](https://www.algolia.com/doc/api-reference/widgets/hierarchical-menu/react-hooks/#widget-param-rootpath) "if the first level is not the root level" so I guess it implies `rootPath` only has one segment

But still, maybe what I should do instead is to `replace` the first occurence of `rootPath`, that would be safer

test/spec/SearchResults/getFacetValues.js
```diff
@@ -407,6 +407,94 @@ test('getFacetValues(hierarchical) returns escaped facet values', function () {
   expect(facetValues).toEqual(expected);
 });
 
+test('getFacetValues(hierarchical) ignores `rootPath`', function () {
```
dhayab, line 410: I might be wrong, but the title feels misleading. `getFacetValues('type')` seems on the contrary to take into account `rootPath`, as the expected data only returns facet values within `cats`.

src/SearchResults/index.js
Haroenv: yes, I think we should enforce that rootPath is at the beginning

test/spec/SearchResults/getFacetValues.js
Haroenv: good point, probably rootPath should only be taken in account for isRefined, not for anything else

src/SearchResults/index.js
aymeric-giraudet: Yeah I added a condition with `startsWith`

src/SearchResults/index.js
```diff
@@ -714,9 +714,20 @@ function extractNormalizedFacetValues(results, attribute) {
...
       results._state.getHierarchicalFacetByName(attribute);
-    var currentRefinementSplit = unescapeFacetValue(
+    var separator =
+      results._state._getHierarchicalFacetSeparator(hierarchicalFacet);
+    var currentRefinement = unescapeFacetValue(
       results._state.getHierarchicalRefinement(attribute)[0] || ''
-    ).split(results._state._getHierarchicalFacetSeparator(hierarchicalFacet));
+    );
+
+    if (currentRefinement.startsWith(hierarchicalFacet.rootPath)) {
```
dhayab, line 723: I think we need to keep to keep it ES5 compliant.

```suggestion
    if (currentRefinement.indexOf(hierarchicalFacet.rootPath) === 0) {
```

Haroenv: Could we instead maybe do the replace without condition? I think it's being compared twice now already. 

Not sure if it's more efficient, but we could also make the first argument of the replace with a regex `new RegExp('^' + ....`. Although now I'm thinking of it, that may be slower

aymeric-giraudet: Yeah I opted for Dhaya's suggestion, it's also clearer for sure

dhayab (approved):

Haroenv (approved):

